PR #17, opened 2019-03-20
Simplify template

zhngx:
Remove office-js-helpers.
Remove jquery.
Remove office-ui-fabric-js.
Add fabric-core stylesheet.
Modify the html and css to make the template look like the old template.

### DIFF
--- a/src/taskpane/taskpane.html
+++ b/src/taskpane/taskpane.html
@@ -20,7 +20,7 @@
     <link rel="stylesheet" href="https://static2.sharepointonline.com/files/fabric/office-ui-fabric-core/9.6.1/css/fabric.min.css"/>
 
     <!-- Template styles -->
-    <link href="src/taskpane/taskpane.css" rel="stylesheet" type="text/css" />
+    <link href="taskpane.css" rel="stylesheet" type="text/css" />
 </head>
 
 <body class="ms-font-m ms-welcome ms-Fabric">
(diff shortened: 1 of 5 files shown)


src/taskpane/taskpane.html
```diff
@@ -14,53 +14,44 @@
     <script type="text/javascript" src="https://appsforoffice.microsoft.com/lib/1.1/hosted/office.js"></script>
 
     <!-- LOCAL -->
```
akrantz: You can delete the comment for the local section now, right?

src/taskpane/taskpane.html
```diff
@@ -14,53 +14,44 @@
...
-    <link rel="stylesheet" href="node_modules/office-ui-fabric-js/dist/css/fabric.min.css" />
-    <link rel="stylesheet" href="node_modules/office-ui-fabric-js/dist/css/fabric.components.css" />
-
+    
     <!-- CDN -->
     <!-- For the Office UI Fabric, go to http://aka.ms/office-ui-fabric to learn more. -->
```
akrantz: We need to replace the comment about learning Office UI Fabric to one that recommends using Fabric React for your add-in to look like Office, or you can replace the HTML/CSS with your own design.

src/taskpane/taskpane.html
akrantz: @zhngx I pushed a commit to fix this.

src/taskpane/taskpane.html
akrantz: We'll probably need to add something here in the future, but this will suffice for now.

PR #19, opened 2019-03-22
devcert integration

asachin96:


package.json
```diff
@@ -50,7 +50,7 @@
     "source-map-loader": "^0.2.4",
     "ts-loader": "^5.3.1",
     "typescript": "^3.3.3",
-    "webpack": "^4.29.5",
+    "webpack": "^4.29.6",
```
akrantz: Was an update needed for webpack?

webpack.config.js
```diff
@@ -1,10 +1,12 @@
+const devCerts = require("office-addin-dev-certs");
 const CleanWebpackPlugin = require("clean-webpack-plugin");
 const CopyWebpackPlugin = require("copy-webpack-plugin");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
 const fs = require("fs");
 const webpack = require("webpack");
 
-module.exports = (env, options) => {
+module.exports = async (env, options) => {
+  const httsServerInfo = await devCerts.gethttpsServerOptions();
```
akrantz: In https://github.com/OfficeDev/Office-Addin-Scripts, this function name should be updated to have an uppercase h so it is `getHttpsServerOptions()`.

akrantz: You should take care to review changes so things like misspellings are caught. In this case, I would rename the variable to `https` which will match the object name.

akrantz:
Now that office-addin-dev-certs is published, you can add it as a dev dependency to the package.

package.json
asachin96: npm command might have updated this

PR #24, opened 2019-04-26
Simplify resource ids

akrantz:
Remove "Contoso." prefix from the resource ids for simplicity and to avoid illustrating a pattern where developers might feel that it is needed and could also easily go past the short string length limit of 32 characters.

I've also renamed the icon and commands group resource ids for clarity and consistency.


akrantz:
@TCourtneyOwen @kbrandl @JuaneloJuanelo I've made these changes to simplify the resource ids.

PR #30, opened 2019-05-02
Provide an add-in command which executes a JavaScript function.

akrantz:
The commands.ts code is updated and adds the function to global scope so it can be called.

### DIFF
--- a/manifest.outlook.xml
+++ b/manifest.outlook.xml
@@ -66,6 +66,21 @@
                     <SourceLocation resid="Taskpane.Url" />
                   </Action>
                 </Control>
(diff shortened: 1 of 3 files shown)


akrantz:
@davidchesnut I am not sure whether these changes will require some doc updates. Per Kim's email, she was updating the tutorial for this and was having problems getting it to work. These changes will unblock that because the code now adds the JavaScript function to global scope so it can be called.

davidchesnut:
@akrantz  I'll check in Kim's branch to be sure the docs are in sync with these changes. Thanks!

akrantz:
@davidchesnut So does that mean I am OK to merge these changes?

PR #33, opened 2019-05-09
VSO. 3341632: Code to make template target a single host needs to be in the templates themselves

TCourtneyOwen:
This moves the convert to single host code out of generator-office into the project itself.  As part of yo office project generation, yo office will set environment variables for projectFolder, host and scriptType and then does the following to invoke the 'convert-to-single-host' npm script in the generated project folder:

              const cmdLine = `npm run convert-to-single-host --prefix ${this.destinationPath()}`;
              await childProcess.exec(cmdLine, (err) => {
                if (err) {
                  reject(err);
                }
              });  

I tested this out in yo office locally but update the taskpane project in projectProperties.json to use this branch when generating a taskpane typescript project and all works well.  The only thing that convertToSingleHost.js doesn't do is delete itself after it runs and I am not sure if this is even possible


TCourtneyOwen:
I think we should add a test to yo office to ensure the generated project is indeed host specific after this change is incorporated

helpers/convertToSingleHost.js
```diff
@@ -0,0 +1,99 @@
...
+    "outlook",
+    "powerpoint",
+    "project",
+    "word"
+];
+
+modifyProjectForSingleHost(projectFolder, host, typescript);
+
+async function modifyProjectForSingleHost(projectFolder, host, typescript) {
+    return new Promise(async (resolve, reject) => {
```
akrantz: You don't need to create a promise for the async function. By definition it returns a promise -- and you should add `: Promise<void>` return type. You can remove the new Promise() and the try/catch.

helpers/convertToSingleHost.js
```diff
@@ -0,0 +1,99 @@
+const fs = require('fs');
+const path = require('path');
+const projectFolder = process.env['projectFolder'];
```
akrantz: You could pass npm arguments. Just use ` -- ` between the script name and the arguments.

`npm run convertToSingleHost -- excel typescript`

helpers/convertToSingleHost.js
```diff
@@ -0,0 +1,99 @@
...
+        }
+    });
+}
+
+async function convertProjectToSingleHost(projectFolder, host, typescript) {
+    try {
+        let extension = typescript ? "ts" : "js";
+
+        // copy host-specific manifest over manifest.xml
+        await fs.readFile(path.resolve(`${projectFolder}/manifest.${host}.xml`), 'utf8', async (err, data) => {
```
akrantz: It would be great to use promise versions of the fs calls and avoid all of the callbacks.

package.json
```diff
@@ -28,7 +28,8 @@
     "unload:powerpoint": "office-toolbox remove -m manifest.xml -a powerpoint",
     "unload:word": "office-toolbox remove -m manifest.xml -a word",
     "validate": "office-toolbox validate -m manifest.xml",
-    "watch": "webpack --mode development --watch"
+    "watch": "webpack --mode development --watch",
+    "convert-to-single-host": "node ./helpers/convertToSingleHost.js"
```
akrantz: Let's put the file into a "scripts" folder (instead of "helpers").

helpers/convertToSingleHost.js
TCourtneyOwen: I seem to recall that the promise versions are still in preview mode, but I'll look into it

akrantz:
@TCourtneyOwen I moved the file out of the scripts folder to the root so that there is no need to delete a folder. I added code to delete the script file itself, and did some other cleanup such as verifying the host parameter, adding a catch handler on the main async function, and removing try/catch that wasn't needed.

convertToSingleHost.js
```diff
@@ -0,0 +1,74 @@
...
+  await writeFileAsync(packageJson, JSON.stringify(content, null, 4));
+}
+
+
+/**
+ * Modify the project so that it only supports a single host.
+ * @param host The host to support.
+ */
+modifyProjectForSingleHost(host).catch(err => {
+    console.error(`Error: ${err instanceof Error ? err.message : err}`);
```
TCourtneyOwen: will this bubble up the error to generator-office?

akrantz: Please check generator-office. What does it do if there is an error?

TCourtneyOwen: It doesn't look like an error bubbles up to yo office.  I set the host param to 'foo' in yo office and it the convert-to-single-host operation did indeed fail, but there was no error reported in yo office output

akrantz:
@TCourtneyOwen The last change here allows the command to print an error message to the user if the convert-to-single-host script has an error, and by setting a non-zero exit code, the child_process.exec() call will be informed that `npm run convert-to-single-host` had an error. I think we are good to merge this, and update the other repositories as well.

convertToSingleHost.js
akrantz: I've update this so that the main function will catch any exception that occurs and display the error. In addition it sets the process.exitCode so that generator-office will know whether the command ran successfully or not. I even tested this in the Windows Command Prompt, calling `echo %errorlevel%` after running both successfully and with an error, and tested with generator-office.

There is an issue that generator-office code doesn't seem to display the error that occurs, and I think a fix can be made there to address it.

package.json
akrantz: I ended up moving into the root to avoid having to delete a folder.

TCourtneyOwen:
I will initiate PRs for the other repos today, making framework adjustments as necessary

PR #35, opened 2019-05-31
update packages

akrantz:


akrantz:
Well it is good to see that the CI loop caught the build error related to the change with CleanWebpackPlugin. I updated the major version, but will need to see what change needs to be made in the weback config.

TCourtneyOwen:
@akrantz  yup I ran into this error yesterday - the latest CleanWebpackPlugin package has a breaking change

akrantz:
The release shows the change needed: https://github.com/johnagan/clean-webpack-plugin/releases/tag/v3.0.0. A named import is needed instead of the default import. I've pushed the fix.

PR #36, opened 2019-06-07
enhance webpack config for https and port

akrantz:
* if '--https false' is specified when calling webpack or webpack-dev-server,
  it will use http instead of https.

  This will run the dev server using http:
  `npm run dev-server -- --https false`

  This will build without prompting for dev certs:
  `npm run build -- --https false`

* Get the port from the package.json config "dev-server-port".

  If the value of this config is changed in package.json, "npm run dev-server" will use it.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -62,9 +62,9 @@ module.exports = async (env, options) => {
     devServer: {
       headers: {
         "Access-Control-Allow-Origin": "*"
-      },
-      https: await devCerts.getHttpsServerOptions(),
-      port: 3000
+      },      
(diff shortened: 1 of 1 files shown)


TCourtneyOwen:
I assume this job is failing because the machine is in a bad state. The machines appear to get in a bad state when a job is kicked off when they are offline. When they subsequently start for reimaging, the job triggers and I believe disrupts artifact deployment, resulting in a machine that’s misconfigured

akrantz:
It just means that I should stop working so late at night! :-)

akrantz:
@TCourtneyOwen I just did "Retry" on the CI pipeline.

akrantz:
And it succeeded.

PR #37, opened 2019-06-07
'npm run build:dev' instead of 'build-dev'

akrantz:


akrantz:
@TCourtneyOwen Mac test has been running for 23 minutes. We'll need a timeout. Need to understand why the test isn't running reliably.

TCourtneyOwen:
The issue was that I had been debugging the issue I was seeing with the React ui-test PR on this machine and Word and Excel were already booted with a breakpoint set in VS code. I killed everything just now on the machine and reran the job and it passes

akrantz:
OK, good. Thanks.

PR #39, opened 2019-06-10
Minor fix to address issue when converting to single host

TCourtneyOwen:
- Without this change, tests will fail for Excel after calling convertToSingleHost with "excel convert-test" params.


TCourtneyOwen:
Note, I already added this fix to Angular and React before the recent merge to master

PR #40, opened 2019-06-17
Fix unregister command for Mac so the manifest is removed

TCourtneyOwen:
- Prior to this change, the manifest was not being properly removed from the wef cache during stopDebugging because the manifestPath wasn't properly reflecting the location of the sideloaded manifest


TCourtneyOwen:
Please disregard this is for testing pruposes only

PR #52, opened 2019-08-02
office-toolbox@0.3.0

akrantz:
office-toolbox@0.3.0 is currently published using the "next" dist-tag. This is to validate that it is OK to publish as latest and update the templates to use it.


akrantz:
The Mac build agent is hanging when running the tests. I was able to validate that the tests run on my Mac.

PR #54, opened 2019-08-16
Update all package versions

TCourtneyOwen:
- This will ensure we pick up important office-addin-scripts changes as well


TCourtneyOwen:
tests are failing due to the timing issue with sideloading and the start of the dev-server. I am working on a local fix for that issue now, but that doesn't preclude these changes from being merged

PR #55, opened 2019-08-19
Add devServer port to startDebugging command

TCourtneyOwen:
- Currently the block of code in runDevServer that waits for the dev server to start is not getting hit because the port is not set. It's possible that this change alone may enable us to work around the problem where tests fail because sideloading occurs before the dev build is complete.  However we still may need to add the new call to the new waitUntilDevBuildIsComplete method that we are adding to office-addin-debugging


TCourtneyOwen:
Tests now pass with this change. However I still think we should consider adding the change to office-addin-debugging to call waitUntilDevBuildIsComplete

akrantz:
@TCourtneyOwen I updated the office-addin-cli version.

PR #73, opened 2020-03-24
Replaces current-processes to address security vulnerability

TCourtneyOwen:
- Also updates other packages to latest versions


TCourtneyOwen:
Please note, looks like our VM deployments were unsuccessful last night so the CI job is not running successfully. Investigating

test/src/test-helpers.ts
```diff
@@ -1,5 +1,5 @@
 import * as childProcess from "child_process";
-import * as cps from "current-processes";
+const find = require('find-process');
```
akrantz: You should use double-quotes for consistency. I don't recall if the eslint/prettier were integrated but you can run prettier to make things consistent.

test/webpack.config.js
```diff
@@ -18,7 +18,8 @@ module.exports = async (env, options) => {
             extensions: [".ts", ".tsx", ".html", ".js"]
         },
         node: {
-            child_process: 'empty'
+            child_process: 'empty',
+            fs: 'empty'
```
akrantz: remind me what this config is for...

TCourtneyOwen:
Regarding the tests not passing in our CI pipeline there appears to be an Azure issue - I cannot even browse to our DevTestLab to investigate what the problem might be with the VMs we deploy. I get an error.  So I am guessing our VM deployment is failing due to some systemic Azure issue

test/webpack.config.js
TCourtneyOwen: I created a separate config for a couple reasons. I don't want to have the additional overhead of building the test files when building the taskpane itself, which would happen if we use a single config file. Also, the tests infra requires the node exclusions for child_process and fs that the regular taskpane doesn't due to the import of the test infrastructure

test/src/test-helpers.ts
TCourtneyOwen: fixed

akrantz: Shouldn't you be using import and not require in TypeScript code?

test/webpack.config.js
akrantz: I didn't mean why is there a webpack config for tests. I meant what is the meaning of the node section in the webpack config and why does fs and child_processes need to be excluded.

test/src/test-helpers.ts
```diff
@@ -1,5 +1,5 @@
 import * as childProcess from "child_process";
-import * as cps from "current-processes";
+const find = require("find-process");
```
akrantz: Shoudln't this be: `import * as findProcess from "find-process" `

test/webpack.config.js
TCourtneyOwen: we get build errors without adding these

test/src/test-helpers.ts
TCourtneyOwen: import doesn't appear to work - I tried that but will let you know the specific error once I get a chance to look at this again

TCourtneyOwen: If I try to simply import it, I get this error:  This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.ts(2497)

akrantz: Did you use this form of import?

`import find = require("find-process")`

akrantz: https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

akrantz: Since there is not a @types/find-process, it would be good to add it, so that the function actually has type info.

There is just one method:

find(type: string, value: number | string | RegExp, strict: bool = false): Promise<Array>

https://github.com/yibn2008/find-process#readme

TCourtneyOwen: I added it locally but it still gave the same error when trying to import. Pushing up another commit that address 100% of Component Governance issues

TCourtneyOwen: OK import find = require("find-process") does in fact work

package.json
```diff
@@ -33,18 +33,19 @@
     "@babel/core": "^7.9.0",
     "@babel/polyfill": "^7.8.7",
     "@babel/preset-env": "^7.9.0",
+    "@types/find-process": "1.2.0",
```
akrantz, line 36: Good. This allows the TypeScript usage to be type-safe.

TCourtneyOwen:
@akrantz how should we proceed with this PR? Unfortunately, we can't get confirmation that this works on Mac because our Mac agents are out of commission and I haven't been able to remote to them

akrantz:
We need to get the mac machines running. Let's talk offline.

TCourtneyOwen:
I repro this as well and suspect the problem lies in this step in the quick start:

To enable HTTPS for your app, create a vue.config.js file in the root folder of the Vue project with the following contents:

JavaScript

Copy
module.exports = {
  devServer: {
    port: 3000,
    https: true
  }
};

I think we need a way to mimic what we do with our other taskpane apps, but I am not sure if we can do that easily. One alternative that works is to add the following to vue.config.js instead of what's currently recommended:

const fs = require("fs");
const path = require("path");
const homedir = require('os').homedir()

module.exports = {
    //...
    devServer: {
      port: 3000,
      https: true,
      key: fs.readFileSync(path.resolve(`${userProfileDir}/.office-addin-dev-certs/localhost.key`)),
      cert: fs.readFileSync(path.resolve(`${userProfileDir}/.office-addin-dev-certs/localhost.crt`)),
      ca: fs.readFileSync(path.resolve(`${userProfileDir}/.office-addin-dev-certs/ca.crt`))
    }
  };

TCourtneyOwen:
Our Mac CI agent is currently down now, and we can't remote to it, so we performed local verification to ensure that install, build and tests pass

PR #81, opened 2020-08-13
ensure production build contains assets

akrantz:
Currently, production builds output the assets files using the hash for the file name, and do not include assets that are in the manifest but not referenced in the code.

Change the webpack config so the output path for assets keeps the original folder path, and add a reference in the code to icons in the manifest so they will also be in the build output.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,8 +5,12 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
 const fs = require("fs");
 const webpack = require("webpack");
 
+const urlDev="https://localhost:3000/";
+const urlProd="https://www.contoso.com/"; // CHANGE THIS TO YOUR PRODUCTION DEPLOYMENT LOCATION
+
 module.exports = async (env, options) => {
   const dev = options.mode === "development";
+  const buildType = dev ? "dev" : "prod";
   const config = {
     devtool: "source-map",
     entry: {
@@ -36,7 +40,10 @@ module.exports = async (env, options) => {
         },
         {
           test: /\.(png|jpg|jpeg|gif)$/,
-          use: "file-loader"
+          loader: "file-loader",
+          options: {
+            name: '[path][name].[ext]',          
+          }
         }
       ]
     },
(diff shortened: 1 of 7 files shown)


TCourtneyOwen:
@akrantz I will try this out tomorrow and add any comments I have

webpack.config.js
```diff
@@ -5,8 +5,12 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
 const fs = require("fs");
```
TCourtneyOwen, line 5: do we want to make an npm command in package.json to run the dev-server for prod?  I don't think we currently have that

webpack.config.js
```diff
@@ -5,8 +5,12 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
...
 const webpack = require("webpack");
 
+const urlDev="https://localhost:3000/";
+const urlProd="https://www.contoso.com/"; // CHANGE THIS TO YOUR PRODUCTION DEPLOYMENT LOCATION
```
TCourtneyOwen, line 9: I am a bit unclear what urlProd is for.  Can you please clarify

webpack.config.js
akrantz: I don't think we need to. It is not common to want to run the dev-server as prod. It can be done with: `npm run dev-server -- --mode production` so I don't think we need a separate npm run script for it.

webpack.config.js
akrantz: yes, when we build for production, the dist output will have a manifest that points to the urlProd location. We can't possibly know what this value should be, the developer would want to specify where they plan to deploy it. I put the fake "https://www.contoso.com/" value so you can see that get written to the manifest.prod.xml in the dist folder.

TCourtneyOwen: Should we perhaps copy the manifest to the dist folder for debug?  This will especially be useful for sideloading to WAC where we will expect the manifest to be served up by the dev-server.  For example by adding the following to webpack.config.js:

      new CopyWebpackPlugin([
        {
          to: "manifest.xml",
          from: "./manifest.xml"
        }
      ]),

This of course can be done in a separate PR but I thought I'd just throw it out there

akrantz: It does copy the manifest for both production and development. The buildType variable is "dev" or "prod" so the manifest output is "manifest.dev.xml" or "manifest.prod.xml". It can handle manifests with other names, e.g. "manifest.excel.xml" will be output as "manifest.excel.dev.xml".

The file doesn't get written out when running the dev server since it serves everything from memory. You can still browse to "https://localhost:3000/manifest.dev.xml". If you do `npm run build -- --mode development` you can also see the dist file output.

PR #82, opened 2020-08-18
"https://" in AppDomain

exextoc:
This to avoid parsing issues when developer adds AppDomains without "https://".

### DIFF
--- a/manifest.outlook.xml
+++ b/manifest.outlook.xml
@@ -14,7 +14,7 @@
   <HighResolutionIconUrl DefaultValue="https://localhost:3000/assets/icon-80.png"/>
   <SupportUrl DefaultValue="https://www.contoso.com/help"/>
   <AppDomains>
-    <AppDomain>contoso.com</AppDomain>
+    <AppDomain>https://www.contoso.com</AppDomain>
   </AppDomains>
   <Hosts>
     <Host Name="Mailbox" />


akrantz:
I'd like to understand better the motivation behind this change. The protocol is not part of the domain. Is this a workaround for a bug in Outlook?

exextoc:
The [documentation](https://docs.microsoft.com/en-us/office/dev/add-ins/outlook/manifests#app-domains) template has always called out `https://www.contoso2.com` as example.
In Office Js also always had a hard check for scheme, `appDomains[i].indexOf("://") !== -1`

    function postDialogMessage(targetWindow, message) {
         var appDomains = OSF._OfficeAppFactory.getInitializationHelper()._appContext._appDomains;
         if (appDomains) {
             for (var i = 0; i < appDomains.length && appDomains[i].indexOf("://") !== -1; i++) {
                  targetWindow.postMessage(message, appDomains[i]);
             }
        }
        ...
    }

Now, we have developers following the generated code and not the documentation and this leads to unexpected add-in behavior.

akrantz:
After investigating this, we should do this for all of the Office hosts, not just Outlook.

akrantz:
I've made the changes in this pull request for all Office hosts: https://github.com/OfficeDev/Office-Addin-TaskPane/pull/82

PR #85, opened 2020-09-06
Support debugging using Edge webview (Chromium or EdgeHTML)

akrantz:
Add launch configs to allow debugging Excel, PowerPoint, or Word desktop apps on Windows when the Edge webview is being used.  

This does not check or enforce for whether the Office desktop apps are using the Edge webview. The developer can run the following command to force it: 

Edge (Legacy version):
`npx office-addin-dev-settings webview manifest.xml edge`

Edge (Chromum version)
`npx office-addin-dev-settings webview manifest.xml edgechromium`


akrantz:
@TCourtneyOwen I added support for Edge Chromium.

.vscode/extensions.json
```diff
@@ -6,6 +6,7 @@
   "recommendations": [
     "msjsdiag.debugger-for-chrome",
     "msjsdiag.debugger-for-edge",
+    "msoffice.microsoft-office-add-in-debugger",
```
TCourtneyOwen, line 9: Does this automatically install the extension if it isn't already?

akrantz: VS Code will prompt to install the recommended extensions when the folder is opened. The user can click the first button to install all of the recommended extensions. The other button allows viewing them in the Extensions pane. They can also close the notification and do nothing.

.vscode/launch.json
```diff
@@ -26,28 +26,73 @@
...
-      "request": "launch",
-      "runtimeExecutable": "npm",
-      "runtimeArgs": ["run", "start", "desktop"],
-      "args": ["--", "--app", "powerpoint"]
+      "name": "Excel Desktop (Edge Legacy)",
+      "type": "office-addin",
+      "request": "attach",
+      "port": 9222,
+      "url": "https://localhost:3000/taskpane.html?_host_Info=Excel$Win32$16.01$en-US$$$$0",
+      "trace": "verbose",
```
akrantz, line 45: We should probably remove "trace" so it isn't present by default.

.vscode/launch.json
```diff
@@ -26,28 +26,73 @@
...
+      "webRoot": "${workspaceRoot}",
+      "timeout": 45000,
```
akrantz, line 47: It might be nice to change this order to match the Chromium launch config. Put timeout after port, then webRoot. For url, it might make sense to put it before port.

akrantz: I've changed the order.

.vscode/launch.json
TCourtneyOwen: I do like the trace flag set for myself because it gives clear indication that the attach process is proceeding. However this is just my personal preference.  Others may feel differently

TCourtneyOwen (approved):

.vscode/launch.json
akrantz: It can be added when extra diagnostic info is desired, but doesn't need to be on by default.

PR #89, opened 2020-11-19
Update debug configs and fix one low vulnerabilty

TCourtneyOwen:


.vscode/launch.json
```diff
@@ -30,8 +30,8 @@
       "type": "edge",
       "request": "attach",
       "useWebView": "advanced",
-      "port": 9222,
-      "timeout": 45000,
+      "port": 9229,
+      "timeout": 60000,
```
akrantz, line 34: This is only 1 minute for the timeout. I would bump this to 10 minutes by adding another zero.

.vscode/launch.json
```diff
@@ -103,9 +103,10 @@
       "preLaunchTask": "Debug: Web"
     },
     {
-      "name": "Office Online (Edge)",
-      "type": "edge",
+      "name": "Office Online (Edge Chromium)",
```
akrantz, line 106: I think you should also keep the original config using "type": "edge" and call it "Office Online (Edge Legacy)".

akrantz (changes requested):

.vscode/launch.json
TCourtneyOwen: So keep the original and add the new one as well?

.vscode/launch.json
TCourtneyOwen: sounds good

.vscode/launch.json
akrantz: Actually I take this back. Since this is the browser and Edge Chromium has replaced the legacy versions (unlike with webviews) I don't think you need the other one.

TCourtneyOwen: OK I will leave the legacy config out

akrantz (approved):

PR #96, opened 2021-01-28
no need to use Promise in async methods

akrantz:
I noticed this after when reviewing content add-in changes. Also, there is a missing await For Word.run() (but not Excel) which I optimized by returning the last promise rather than awaiting it. I wonder if this could help avoid some sleeps in the code since I think this could cause the test to return earlier instead of waiting for the Office.js code to execute.

### DIFF
--- a/test/src/test-helpers.ts
+++ b/test/src/test-helpers.ts
@@ -2,64 +2,53 @@ import * as childProcess from "child_process";
 import find = require("find-process")
 
 export async function closeDesktopApplication(application: string): Promise<boolean> {
-    return new Promise<boolean>(async function (resolve, reject) {
-        let processName: string = "";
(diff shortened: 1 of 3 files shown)


test/src/test-helpers.ts
```diff
@@ -75,20 +64,9 @@ export async function sleep(ms: number): Promise<any> {
...
-                        resolve(process[0].pid);
-                    } else {
-                        resolve(undefined);
-                    }                    
-                });
-        } catch (err) {
-            reject(err);
-        }
-    });
+    const [process] = await find('name', processName, false /* strict */)
```
akrantz, line 67: This is a simpler way to get the first item from the array. If the array is empty, then process will be undefined.

TCourtneyOwen (approved):

PR #95, opened 2021-01-27
Content add in

arttarawork:

PR #97, opened 2021-01-28
Adding PowerPoint Test for Taskpane

arttarawork:
There is also a small change to the test runner to only shut off the test server once the very last host has been tested. Otherwise, there's an issue where the dev server closes and doesn't reopen before the next host gets tested. This issue hasn't been encountered before, since it's only a problem if the previous host wasn't Excel. We only previously tested Excel and Word (in that order), so this includes a small change to accommodate that issue for including another host.

### DIFF
--- a/src/taskpane/powerpoint.ts
+++ b/src/taskpane/powerpoint.ts
@@ -22,15 +22,20 @@ export async function run() {
   /**
    * Insert your PowerPoint code here
    */
-  Office.context.document.setSelectedDataAsync(
-    "Hello World!",
-    {
-      coercionType: Office.CoercionType.Text
-    },
-    result => {
-      if (result.status === Office.AsyncResultStatus.Failed) {
-        console.error(result.error.message);
+
+  return new Promise<void>((resolve, reject) => {
+    Office.context.document.setSelectedDataAsync(
+      "Hello World!",
+      {
+        coercionType: Office.CoercionType.Text
+      },
+      result => {
+        if (result.status === Office.AsyncResultStatus.Failed) {
+          console.error(result.error.message);
+          reject(result.error);
+        }
+        resolve();
       }
-    }
-  );
+    );
+  });
 }
--- a/test/src/powerpoint-test-taskpane.ts
+++ b/test/src/powerpoint-test-taskpane.ts
@@ -1,0 +1,34 @@
+import { pingTestServer, sendTestResults } from "office-addin-test-helpers";
+import { run } from "../../src/taskpane/powerpoint";
+import * as testHelpers from "./test-helpers";
+const port: number = 4201;
+let testValues: any = [];
+
+Office.onReady(async (info) => {
+    if (info.host === Office.HostType.PowerPoint) {
+        const testServerResponse: object = await pingTestServer(port);
+        if (testServerResponse["status"] == 200) {
+            runTest();
+        }
+    }
+});
+
+export async function runTest(): Promise<void> {
+    // Execute taskpane code
+    await run();
+    await run();
+    await testHelpers.sleep(6000);
+
+    // Get output of executed taskpane code
+    return Office.context.document.getSelectedDataAsync(Office.CoercionType.Text, async (asyncResult) => {
+        if (asyncResult.status === Office.AsyncResultStatus.Failed) {
+            console.error(asyncResult.error.message);
+            testHelpers.addTestResult(testValues, "output-message", asyncResult.error.message, " Hello World!");
+        } else {
+            console.log(`The selected data is "${asyncResult.value}".`);
+            testHelpers.addTestResult(testValues, "output-message", asyncResult.value, " Hello World!");
+        }
+        await sendTestResults(testValues, port);
+        testValues.pop();
+    });
+}
--- a/test/src/powerpoint-test-taskpane.ts
+++ b/test/src/powerpoint-test-taskpane.ts
@@ -14,21 +14,43 @@ Office.onReady(async (info) => {
 });
 
 export async function runTest(): Promise<void> {
+    // Set up textbox and cursor for taskpane code test
+    await new Promise<void>((resolve, reject) => {
+        Office.context.document.setSelectedDataAsync(
+            " ",
+            {
+                coercionType: Office.CoercionType.Text
+            },
+            result => {
+                if (result.status === Office.AsyncResultStatus.Failed) {
+                    console.error(result.error.message);
+                    reject(result.error);
+                }
+                resolve();
+            }
+        )
+    });
+
     // Execute taskpane code
     await run();
-    await run();
     await testHelpers.sleep(6000);
+    await actualTest();
+}
 
+async function actualTest() {
     // Get output of executed taskpane code
-    return Office.context.document.getSelectedDataAsync(Office.CoercionType.Text, async (asyncResult) => {
-        if (asyncResult.status === Office.AsyncResultStatus.Failed) {
-            console.error(asyncResult.error.message);
-            testHelpers.addTestResult(testValues, "output-message", asyncResult.error.message, " Hello World!");
-        } else {
-            console.log(`The selected data is "${asyncResult.value}".`);
-            testHelpers.addTestResult(testValues, "output-message", asyncResult.value, " Hello World!");
-        }
-        await sendTestResults(testValues, port);
-        testValues.pop();
+    return new Promise<void>((resolve) => {
+        Office.context.document.getSelectedDataAsync(Office.CoercionType.Text, async (asyncResult) => {
+            if (asyncResult.status === Office.AsyncResultStatus.Failed) {
+                console.error(asyncResult.error.message);
+                testHelpers.addTestResult(testValues, "output-message", asyncResult.error.message, "Hello World!");
+            } else {
+                console.log(`The selected data is "${asyncResult.value}".`);
+                testHelpers.addTestResult(testValues, "output-message", asyncResult.value, "Hello World!");
+            }
+            await sendTestResults(testValues, port);
+            testValues.pop();
+            resolve();
+        });
     });
 }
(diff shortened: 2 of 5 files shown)


test/src/powerpoint-test-taskpane.ts
```diff
@@ -0,0 +1,34 @@
...
+        if (testServerResponse["status"] == 200) {
+            runTest();
+        }
+    }
+});
+
+export async function runTest(): Promise<void> {
+    // Execute taskpane code
+    await run();
+    await run();
```
akrantz, line 19: only need one line that has `await run();`

test/src/test-taskpane.ts
```diff
@@ -1,2 +1,3 @@
 import "./excel-test-taskpane";
-import "./word-test-taskpane"; 
+import "./word-test-taskpane";
```
akrantz, line 2: please put in alphabetic order

test/ui-test.ts
```diff
@@ -7,7 +7,7 @@ import * as officeAddinTestHelpers from "office-addin-test-helpers";
 import * as officeAddinTestServer from "office-addin-test-server";
 import * as path from "path";
 import * as testHelpers from "./src/test-helpers";
-const hosts = ["Excel", "Word"];
+const hosts = ["Excel", "Word", "PowerPoint"];
```
akrantz, line 10: Please put in alphabetic order.

test/ui-test.ts
```diff
@@ -50,7 +50,9 @@ hosts.forEach(function (host) {
             assert.equal(stopTestServer, true);
 
             // Unregister the add-in
-            await stopDebugging(manifestPath);
+            if (host == hosts[hosts.length - 1]) {
```
akrantz, line 53: Why are you adding this check that there is only one host? This shouldn't be required as far as I am aware.

akrantz: Really I think you are pointing out that this could be optimized to start the dev server once, then run the test for each host (which includes startDebugging() and stopDebugging()), and then it can do after(). I don't think you should need to change this here because it is beyond the task of adding the powerpoint support.

akrantz (changes requested):
Please see the comments.

test/ui-test.ts
arttarawork: Without this change, whatever tests following a non-Excel host will fail. I can change the tests so we start/end the dev server in the before/after, though.

akrantz: The problem with the dev server that you mentioned would require fixing this to start the dev server once before the loop for each host, and then stopping it afterwards. I guess you should go ahead and make this change.

test/src/powerpoint-test-taskpane.ts
arttarawork: https://github.com/OfficeDev/Office-Addin-TaskPane/pull/95#discussion_r565726957

arttarawork: "Weird behavior with getSelectedDataAsync(). Running run() once adds "Hello World!" to the slide, but doesn't select the new text properly, so running getSelectedDataAsync() after running run() once returns "". Running run() twice inserts "Hello World! Hello World!". At this point, the latter part (" Hello World!") is what's selected, which is what getSelectedDataAsync() reads.

It's a workaround I don't like, so if there's a proper way to select that text, I would use it, but I haven't found anything that lets me do that."

akrantz: @wandyezj Could you help see if there a way to update the test code here so it has the selected text? It would help to understand if this highlights a bug with PowerPoint, a timing problem, or something else.

test/ui-test.ts
```diff
@@ -63,6 +63,8 @@ hosts.forEach(function (host) {
     });
 });
 
-after(`Unregister the add-in`, async function () {
-    return stopDebugging(manifestPath);
+after(`Unregister the add-in`, callback => {
```
akrantz, line 66: Do you not like using async/await code? 

I have this vague recollection that we use function() { } for the callback rather than arrow notation () => { } because the latter has some issue with mocha.

```
after("Unregister the add-in", async function(callback) => {
  await stopDebugging(manifestPath);
  setImmediate(callback);
}
```

arttarawork: Gotcha, no I actually prefer the code I had in its place earlier, with just the stopDebugging.

This is mostly because there's a weird issue where the tests are failing on Mac but not on windows. 

I figured maybe this would fix it? https://github.com/sindresorhus/gulp-mocha/issues/129

But it clearly didn't, based off the latest pipeline run.

test/ui-test.ts
TCourtneyOwen: I think you are still start and stopping the dev-server multiple times here, which can lead to test unreliability.  It's something we should have fixed long ago.  You can see an example for how I am doing in this another branch at https://github.com/TCourtneyOwen/Office-Addin-TaskPane/blob/add-web-tests/test/ui-tests.ts

test/src/powerpoint-test-taskpane.ts
akrantz: @TCourtneyOwen How do you feel about this redundant `await run()` as a workaround for this issue? Is there some way to avoid the issue? I feel like it would be better to only have a single run() if possible and would like to understand better the reason for the issue. Should we change the test code to something else to avoid it?

test/src/test-helpers.ts
```diff
@@ -8,7 +8,7 @@ export async function closeDesktopApplication(application: string): Promise<bool
             processName = "Excel";
             break;
         case "powerpoint":
-            processName = (process.platform === "win32") ? "Powerpnt" : "Powerpoint";
+            processName = (process.platform === "win32") ? "Powerpnt" : "PowerPoint";
```
akrantz, line 11: Are all non-win32 platforms "Powerpnt"? Seems like it would be better to use a switch statement with "Powerpoint" as the default case, and the specific platforms called out which need the alternate spelling.

test/src/powerpoint-test-taskpane.ts
TCourtneyOwen: Can't we use setSelectedDataAsync to set the text:  https://docs.microsoft.com/en-us/javascript/api/office/office.document?view=powerpoint-js-1.1#setSelectedDataAsync_data__callback_

arttarawork: The run function in `src\taskpane\powerpoint.ts` is already using setSelectedDataAsync:
```
export async function run() {
  /**
   * Insert your PowerPoint code here
   */
  Office.context.document.setSelectedDataAsync(
    "Hello World!",
    {
      coercionType: Office.CoercionType.Text
    },
    result => {
      if (result.status === Office.AsyncResultStatus.Failed) {
        console.error(result.error.message);
      }
    }
  );
}
```

wandyezj: How come the setSelectedDataAsync is not awaited?

arttarawork: I can put an await on `setSelectedDataAsync`, but the same issue remains. I'll push out a change with the await, if requested, though.

test/ui-test.ts
```diff
@@ -49,14 +49,15 @@ hosts.forEach(function (host) {
             const stopTestServer = await testServer.stopTestServer();
             assert.equal(stopTestServer, true);
 
-            // Unregister the add-in
-            await stopDebugging(manifestPath);
-
             // Close desktop application for all apps but Excel, which has it's own closeWorkbook API
             if (host != 'Excel') {
                 const applicationClosed = await testHelpers.closeDesktopApplication(host);
                 assert.equal(applicationClosed, true);
```
TCourtneyOwen, line 55: Can you please change assert.equal to assert.strictEqual throughout

src/taskpane/powerpoint.ts
```diff
@@ -22,15 +22,20 @@ export async function run() {
...
-  Office.context.document.setSelectedDataAsync(
-    "Hello World!",
-    {
-      coercionType: Office.CoercionType.Text
-    },
-    result => {
-      if (result.status === Office.AsyncResultStatus.Failed) {
-        console.error(result.error.message);
+
+  return new Promise<void>((resolve, reject) => {
```
akrantz, line 26: We prefer to use async/await where feasible. This looks to do the same thing as before except where the `reject(result.error)` is which should be `throw result.error` (instead of `console.error()`) in the original async code. Is that really the fix?

test/src/powerpoint-test-taskpane.ts
```diff
@@ -14,21 +14,43 @@ Office.onReady(async (info) => {
 });
 
 export async function runTest(): Promise<void> {
+    // Set up textbox and cursor for taskpane code test
+    await new Promise<void>((resolve, reject) => {
```
akrantz, line 18: this is doing the same thing as the redundant `run()`. This isn't any better.

test/src/powerpoint-test-taskpane.ts
```diff
@@ -14,21 +14,43 @@ Office.onReady(async (info) => {
...
+                resolve();
+            }
+        )
+    });
+
     // Execute taskpane code
     await run();
-    await run();
     await testHelpers.sleep(6000);
+    await actualTest();
```
akrantz, line 37: I am confused why this was added. We want to test the `run()` method of the original code. Adding the actualTest() function doesn't help with that.

arttarawork: The actualTest method puts the code that reads the result separately for readability. I can add it back to runTest if needed.

test/src/powerpoint-test-taskpane.ts
akrantz: This has now been addressed.

test/ui-test.ts
arttarawork: Done!

test/src/test-helpers.ts
arttarawork: Win32 uses Powerpnt, while Mac uses PowerPoint. I don't think we run on any other native operating systems.

akrantz: ok fine.

akrantz (approved):

PR #98, opened 2021-01-29
update packages

akrantz:


akrantz:
`npm outdated --long` will show which packages are outdated.

I used `npm update --dev` to update all packages (respecting semver), which should be safe (no breaking changes).

For `ts-node`, there is a major version update because it deprecated of Node versions < 10. This should not have an impact and running the tests should validate that this is OK.

The only packages not updated to the latest available version is webpack and related packages. Webpack 5 is a major update which breaks a lot of things and will require significant work to update.

package.json
```diff
@@ -30,38 +30,38 @@
...
-    "mocha": "^8.1.3",
-    "office-addin-cli": "^1.0.13",
-    "office-addin-debugging": "^3.0.34",
-    "office-addin-dev-certs": "^1.5.5",
-    "office-addin-lint": "^1.0.26",
-    "office-addin-manifest": "^1.5.12",
+    "html-webpack-plugin": "^4.5.1",
+    "mocha": "^8.2.1",
+    "office-addin-cli": "^1.0.16",
+    "office-addin-debugging": "^3.0.42",
```
TCourtneyOwen, line 51: Can we wait until we include the major version update change for office-addin-debugging? Also, I think we made appType required for sideload in office-addin-dey-settings and this should also be a major version change.  As soon as we get Joe's recent PR merged to fix startDebugging, I can publish new versions of the packages and we can update the package versions here

akrantz: We shouldn't make appType required. We should just default to AppType.Desktop if not provided.

I would suggest that after merging the office-addin-scripts PR, we publish to the "next" dist-tag, and then you can use this branch to update have another commit to update to the newer packages. If that all looks good, then release the office-addin-scripts dist-tag from next to latest, and then we can merge this PR.

TCourtneyOwen: So I better fire up another PR to make appType unrequired. I recall we discussed this at one point and decided it should be a required cli param for the sideload command, but I can't remember offhand why we thought that

akrantz: Thanks. Sorry if there was a miscommunication but I think it was meant that we fall back to the default (versus prompting), and not that we require the function to provide it. Since it has not a required parameter now, there is no need to change it to be required and it's more convenient if it isn't because we can still choose to change the behavior if not specified by the caller.

akrantz:
Merged into https://github.com/OfficeDev/Office-Addin-TaskPane/pull/101

PR #100, opened 2021-02-10
Enable debugging/sideload for Outlook (desktop and web)

TCourtneyOwen:
- Add debug configs for Outlook
- Update office-addin-debugging version to the one that supports outlook and web sideloading
- Add new "start:desktop:outlook" script
- Update convertToSingleHost to remove "start:desktop:outlook"


igor-ribeiiro:
The code looks fine to me, just have a look into the failing check case to understand what happened

TCourtneyOwen:
I need to update the tests to use the new method signature for startDebugging

igor-ribeiiro (approved):
Looks good for me!

convertToSingleHost.js
```diff
@@ -88,6 +88,7 @@ async function updatePackageJsonForSingleHost(host) {
     if (key.startsWith("sideload:")
       || key.startsWith("unload:")
       || key === "convert-to-single-host"
+      || key === "start:desktop:outlook"
```
akrantz, line 91: Why does this get deleted?

.vscode/launch.json
```diff
@@ -47,6 +47,28 @@
...
+      "timeout": 600000,
+      "webRoot": "${workspaceRoot}",
+      "preLaunchTask": "Debug: Outlook Desktop",
+      "postDebugTask": "Stop Debug"
+    },
+    {
+      "name": "Outlook Desktop (Edge Legacy)",
+      "type": "office-addin",
+      "request": "attach",
+      "url": "https://localhost:3000/taskpane.html?_host_Info=Outlook$Win32$16.01$en-US$$$$0",
```
akrantz, line 65: Did you ever update the vscode extension so the url could be optional?  Just asking, no change needed.

akrantz (approved):
Looks OK overall. See comments.

package.json
```diff
@@ -7,9 +7,10 @@
   },
   "license": "MIT",
   "config": {
-    "app_to_debug": "excel",
-    "app_type_to_debug": "desktop",
-    "dev_server_port": 3000
+    "app-to-debug": "excel",
```
akrantz, line 10: This is reverting some changes that were made on master.

package.json
```diff
@@ -7,9 +7,10 @@
...
+    "app-type-to-debug": "desktop",
+    "dev-server-port": 3000,
+    "document": ""
```
akrantz, line 13: Why a blank document? Is this just so someone can see where to add it? Personally, I think we should leave "document" off and let them use the -d or --document parameter. We can document how to add this key to package.json to make it the default, but I don't think the empty value helps too much. 

Also, does this apply for both web and desktop sideload?  Have we really tested this out in practice? I think it really should have been two separate values that were more descriptive, e.g. "document_for_sideload_to_web" and "document_for_sideload_to_desktop".  Might need to chat more about this.

package.json
```diff
@@ -22,6 +23,7 @@
     "prettier": "office-addin-lint prettier",
     "start": "office-addin-debugging start manifest.xml",
     "start:desktop": "office-addin-debugging start manifest.xml desktop",
+    "start:desktop:outlook": "office-addin-debugging start manifest.outlook.xml desktop --app outlook",
```
akrantz, line 26: why are we doing this just for outlook? we didn't do it for other apps.

test/ui-test.ts
```diff
@@ -27,8 +27,8 @@ hosts.forEach(function (host) {
             // Call startDebugging to start dev-server and sideload
             const devServerCmd = `npm run dev-server -- --config ./test/webpack.config.js `;
             const devServerPort = parseNumber(process.env.npm_package_config_dev_server_port || 3000);
-            await startDebugging(manifestPath, AppType.Desktop, toOfficeApp(host), undefined, undefined, 
-                devServerCmd, devServerPort, undefined, undefined, undefined, false /* enableDebugging */);
+            await startDebugging(manifestPath, { appType:AppType.Desktop, app:toOfficeApp(host), devServerCommandLine:devServerCmd,
```
akrantz, line 30: Please fix formatting. There should be a space after the colons.

akrantz (changes requested):
See comments. (Sorry forgot to start a review first.)

convertToSingleHost.js
TCourtneyOwen: We don't want this to remain in the yo office generated project - we just want the standard start:deskop one in the generated project which will work because manifest.outlook.xml will be renamed to manifest.xml.

akrantz:
@TCourtneyOwen What is the status with this pull request? Are you planning to address the comments?

package.json
TCourtneyOwen: We could leave it off - I am not adamant about including it.  You can include a desktop document too for sideloading but it will only work if that document already contains a webextension part.  Michelle documented sideloading to web and it tells how to use the "document" param in the context of a project. We could just leave this out of the template and people can add it if they want

package.json
TCourtneyOwen: This script is only intended for internal use and will be stripped out during project generation.  The other hosts don't need it because they all use manifest.xml and can be invoked via the "start" script

TCourtneyOwen:
@akrantz I have been holding off on this PR until we can get the Outlook sideloading flight enabled for Prod - this is taking longer than I would like due to having to wait for PR approvals

package.json
akrantz: @TCourtneyOwen Please fix this. Thanks. It needs to use underscores.

package.json
akrantz: Let's leave `document` off and let people add it if they want.

package.json
akrantz: The script isn't for internal use, but is only needed when the template contains multiple hosts as outlook has a different manifest. When converted to a single host, manifest.outlook.xml is renamed to manifest.xml and then there isn't a need for a outlook-specific start command.

This is OK. I was just missing the context that manifest.outlook.xml gets renamed.

akrantz:
There are still outstanding comments that you could address. 

Even after the flight is enabled in Prod there may be some folks that don't have it and sideload won't work. Could we add a check to determine if Outlook can sideload and put an error message if not?

akrantz (changes requested):
See comments.

.vscode/launch.json
TCourtneyOwen: I did not

package.json
TCourtneyOwen: Removing "document" from template

package.json
TCourtneyOwen: fixed locally

TCourtneyOwen:
Regarding this comment, are you suggesting a Office version check:

Even after the flight is enabled in Prod there may be some folks that don't have it and sideload won't work. Could we add a check to determine if Outlook can sideload and put an error message if not?

If so, I think we would need to add that to office-addin-dev-settings

akrantz:
Yes, I don't think the scripts can check whether the flight is enabled, but you could at least validate that the Office version has support for sideloading.

TCourtneyOwen:
I will see if I can find a way to determine the Outlook install version from the registry and we can add a check to office-addin-dev-settings.  I did complete the flight yesterday to roll out to Prod and verified that the feature works with a Prod version of Outlook, so that's cool

akrantz:
That's great. Thanks.

TCourtneyOwen:
I found the installed version for C2R installs under HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Office\ClickToRun\Configuration in the VersionToReport entry, but I am not sure if there is a similar key for MSI installs

akrantz:
@TCourtneyOwen Perhaps you should work to land this PR without the version check, and a separate PR can be done to add that. Perhaps Igor can work on it.

akrantz (approved):

TCourtneyOwen:
@akrantz I think that's a great idea and in fact I was thinking of asking Igor to do the version check as well.

@igor-ribeiiro I will create a work item for you to check the Outlook install version as part of the sideload command to determine if it's a valid version (needs to >= 16.0.13709.10000). We can check the installed version using our existing registry methods in Office-Addin-Scripts to check the ClientVersionToReport setting under HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Office\ClickToRun\Configuration

PR #103, opened 2021-02-24
BugFix: Making debugging web via vscode work again

igor-ribeiiro:
Tested by running the debug command from VSCode interface and it correctly opens and loads the add-in for debugging in the web.

**That was originally the bug:**
VScode prompts for the url and this url is not passed to the ```start:web``` command.

Steps to reproduce:
1. Create a new web office document
2. Copy the shared link url of the document
3. Go the vs code, open debug panel and run Office Online (can be Chrome or Edge)
4. When it prompts for the url, paste the shared link
Result:
```Debugging is being started...
App type: web
Sideloading the Office Add-in...
Error: Unable to start debugging.
office-addin-debugging: Unable to sideload the Office Add-in.  
Error: For sideload to web, you need to specify a document url.```

Expect:
It should sideload successfully and open the page

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -95,10 +95,6 @@
       "name": "Office Online (Chrome)",
       "type": "chrome",
       "request": "launch",
-      // To debug your Add-in:
-      // 1. When prompted, enter the url (share link) to an Office Online document.
-      // 2. Sideload your Add-in. https://docs.microsoft.com/en-us/office/dev/add-ins/testing/sideload-office-add-ins-for-testing
-      "url": "${input:officeOnlineDocumentUrl}",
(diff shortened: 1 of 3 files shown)


TCourtneyOwen:
I think this looks fine - just need to see why the test is hanging for the CI loop

igor-ribeiiro:
> I think this looks fine - just need to see why the test is hanging for the CI loop

How can I do that?

TCourtneyOwen:
This change wouldn't have impacted tests at all, but I think the problem is that the PPT test is hanging because the taskpane is not opening properly

TCourtneyOwen (approved):

akrantz (approved):

PR #104, opened 2021-03-05
Remove delete of .gitignore from converToSingleHost

TCourtneyOwen:
- .gitignore doesn't get downloaded with the zip file when we create a project.  convertToSingleHost fails as a result and subquent install process doesn't proceed


millerds (dismissed):

TCourtneyOwen:
Something is definitely not right with our linter. It's complaining about errors such as the following:

Delete `␍`  prettier/prettier

I see several articles about how to fix this, but wonder this error suddenly cropped up

PR #106, opened 2021-03-09
Fixing linting error

igor-ribeiiro:
```npm run lint``` was giving errors. So I ran ```npm run prettier``` and ```npm run lint:fix``` to solve.

Now there are no lint issues

Tests:
- Tested with ```npm run test```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,473 +5,452 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.12.13",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.13.tgz",
-      "integrity": "sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==",
+      "version": "7.10.4",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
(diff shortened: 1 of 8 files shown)


akrantz:
You don't need the package-lock.json changes. The other changes are fine.

igor-ribeiiro:
> You don't need the package-lock.json changes. The other changes are fine.

Why is not better for me to update the package-lock.json file?? It simply goes and updates the packages versions, so it looks good for it to be updated

akrantz:
It's not related to what you are fixing. It's better for a PR to be targeted to a specific thing. 

Also, packages are updated all the time, and the point of the package-lock file is to stay on consistent versions until a change to the package dependencies are needed, typically when you install, update, or remove a package. We need to make package updates periodically, but it would be better for this to be done separately. Ideally, we would automate this.

igor-ribeiiro:
> It's not related to what you are fixing. It's better for a PR to be targeted to a specific thing.
> 
> Also, packages are updated all the time, and the point of the package-lock file is to stay on consistent versions until a change to the package dependencies are needed, typically when you install, update, or remove a package. We need to make package updates periodically, but it would be better for this to be done separately. Ideally, we would automate this.

Sure, I just reverted the change

akrantz:
When I checkout your branch and run prettier, it undoes your changes here. Are we using different rules somehow?

igor-ribeiiro:
> When I checkout your branch and run prettier, it undoes your changes here. Are we using different rules somehow?

That's weird. I literally ran this command with the repository fresh (with my virtual desktop). 
But anyway the CI tests are not passing, as if no change was made. I will check that problem still.

PR #111, opened 2021-03-23
Reveting rules change

igor-ribeiiro:
Reverting last changes to .eslintrc.json, given that they are not needed

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,9 +11,6 @@
     },
     "project": "./tsconfig.json"
     },
-  "rules": {
-    "prettier/prettier": ["error", { "endOfLine": "auto" }]
-  },
   "extends": [
     "eslint-config-office-addins"
   ]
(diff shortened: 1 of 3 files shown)


akrantz (dismissed):
@TCourtneyOwen you should pick this up and merge to the other templates.

igor-ribeiiro:
The error is still happening because it is not using the latest version of office-addin-lint yet. It was only pushed one day ago and not published yet
Should I publish it so it can use that??

akrantz (changes requested):
Run `npm outdated --long`. You'll notice that `eslint-config-office-addins` also needs to be updated. This is what actually has the defaults for the `endOfLine: auto` change, right?

igor-ribeiiro:
> Run `npm outdated --long`. You'll notice that `eslint-config-office-addins` also needs to be updated. This is what actually has the defaults for the `endOfLine: auto` change, right?

The changes should be in Office-Addin-Lint as I just checked, but I updated the packaged anyway

akrantz (approved):

PR #117, opened 2021-05-12
Switch to use lint config from eslint-config-office-addins to eslint-plugin-office-addins

millerds:
Currently this project takes linting configuration from eslint-config-office-addins and duplicate some of it directly.  This change will switch to use eslint-plugin-office-addins and remove the extra config lines that can be inherited.


akrantz (approved):

Aniruddha-Shriwant:
Hey, @akrantz & @millerds I was having a naive doubt I apologize for it in advance :)

Can you tell me which check is this?
image

I was curious about this, is it something related to the testing purpose? 
I was working on this project: https://github.com/accordproject/cicero-word-add-in
So I'm having a doubt related to testing, is there a way to test the add-in with some PR checks...
Like usually when we have a normal codebase and suppose a person opens a PR to the repo, then we can have build check deploy checks and maybe some linting checks...But in the case of Add-in if suppose a user created a PR making some changes to code how we can test the add-in currently we just have to clone the user's branch locally and do the testing is there any other way to do it?

millerds:
@Aniruddha-Shriwant that check runs our test suit (seen inside this repo) after building and running the linter on a few different environments we've made for internal use.  It's the same tests you'd see executed when running 'npm run test' in this repo.  Note that these are tests specific to this template repo and so they wouldn't necessarily work in your case, but you can get an idea of what we're doing.  I'm not the one who setup the CI pipeline so I couldn't explain that part of it.

Aniruddha-Shriwant:
Ohhh! Thanks for that information @millerds :)

package.json
```diff
@@ -48,7 +48,7 @@
     "babel-loader": "^8.2.2",
     "clean-webpack-plugin": "^3.0.0",
     "copy-webpack-plugin": "^6.4.1",
-    "eslint-config-office-addins": "^1.0.22",
+    "eslint-plugin-office-addins": "^0.2.0",
```
igor-ribeiiro, line 51: Why is this downgrade happening?

millerds: It's not just a downgrade . . . it's a change in the package being referenced.

igor-ribeiiro (approved):

PR #118, opened 2021-05-17
Fix convert script host update

millerds:
Property name was change in package.json and the convertToSingleHost script needed a matching update.


convertToSingleHost.js
```diff
@@ -76,7 +76,7 @@ async function updatePackageJsonForSingleHost(host) {
   let content = JSON.parse(data);
 
   // update 'config' section in package.json to use selected host
-  content.config["app-to-debug"] = host;
```
igor-ribeiiro, line 79: I think we have a convention to use this hyphen instead of the underscore.
@akrantz , can you see this change? I remember us having a similar discussion about this in one of the .json scripts.

millerds: @TCourtneyOwen changed the package.json file from using hyphens to underscores just 3 months ago.  Not sure the specific reasoning behind it.

akrantz: Because NPM changed. We would have preferred to use hyphen but reading the config values no longer converted the hypen to underscores for the NPM env var. The fix was to change these to use underscores.

millerds: So we stick with the underscores.

igor-ribeiiro (approved):

PR #124, opened 2021-07-07
Update template scripts versions

millerds:
Update the template to use current version of the packages in the office-addin-scripts repo and fix some vulnerabilities.


package.json
```diff
@@ -23,6 +23,7 @@
     "dev-server": "webpack-dev-server --mode development",
     "lint": "office-addin-lint check",
     "lint:fix": "office-addin-lint fix",
+    "lint:test": "office-addin-lint check --test",
```
akrantz, line 26: what is does --test param do?

package.json
```diff
@@ -48,20 +49,21 @@
     "babel-loader": "^8.2.2",
     "clean-webpack-plugin": "^3.0.0",
     "copy-webpack-plugin": "^6.4.1",
-    "eslint-plugin-office-addins": "^0.2.2",
+    "eslint": "^7.29.0",
```
akrantz, line 52: Is it intentional to add eslint as a direct dependency now?

test/webpack.config.js
```diff
@@ -4,76 +4,79 @@ const devCerts = require("office-addin-dev-certs");
...
+const path = require("path");
 
 module.exports = async (env, options) => {
-    // const dev = options.mode === "development";
-    const config = {
-        devtool: "source-map",
-        entry: {
-            polyfill: ["core-js/stable", "regenerator-runtime/runtime"],
-            test: path.resolve(__dirname, './src/test-taskpane.ts')
+  // const dev = options.mode === "development";
```
akrantz, line 10: Are these changes from running prettier on this file?

package.json
millerds: It used the test config of the linter rules instead of the recommended one.  The test config is where the new linter rules we are making get added before we make them standard.  This allows us to run those rules as a test case.

akrantz: You shouldn't add this script to the template. Just use: `npm run lint -- --test` to try it out.

package.json
millerds: It was being flagged as a peer dependency that was needed so I added it.  Not that office-addins-lint is good for a command line, but doesn't work for the live linting of files.

test/webpack.config.js
millerds: Yes . . . running prettier on this file (and the others) wanted all the commas and double quotes.

package.json
millerds: That wasn't always working for me . . . but sure, I can remove it.

package.json
akrantz: The original intention was not to need the direct eslint dependency. What happens if office-addin-lint and this project have conflicting eslint versions? What package is declaring the peer dependency that is an issue?

test/webpack.config.js
akrantz: That's fine. It's just hard to tell whether there were any other changes, but I assume not.

package.json
akrantz: You have to use -- before adding other arguments passed to the underlying command in the script. Otherwise, they are passed to npm instead.

package.json
millerds: Supposedly the advantage of plugins specifying peer dependencies instead of direct dependencies is that it allows them to be resolve into one case, but I can't say that I understand it completely yet.

Looks like I undid what ever had flagged it because removing it doesn't cause and error.  I'm pulling it back out.

akrantz (approved):

PR #125, opened 2021-07-22
Update scripts packages

millerds:
Includes updated linter rules for bug bash


test/ui-test.ts
```diff
@@ -60,11 +60,8 @@ hosts.forEach(function(host) {
       const stopTestServer = await testServer.stopTestServer();
       assert.strictEqual(stopTestServer, true);
 
-      // Close desktop application for all apps but Excel, which has it's own closeWorkbook API
-      if (host != "Excel") {
-        const applicationClosed = await testHelpers.closeDesktopApplication(host);
-        assert.strictEqual(applicationClosed, true);
-      }
+      const applicationClosed = await testHelpers.closeDesktopApplication(host);
```
igor-ribeiiro, line 63: Is this working for Excel?
I remember using a different command a few months back in the CF template to close excel

igor-ribeiiro (approved):

test/ui-test.ts
millerds: In CF you had to make the shared function . . . here it already existed and was being filtered out for Excel, but shouldn't have been.

PR #128, opened 2021-08-20
Update webpack v5

millerds:
There is a vulnerability reported in webpack v4.  We need to update to v5 and take care of the various breaking changes that occurred as a result.

The full webpack provided migration document is at https://webpack.js.org/migrate/5/.

The main changes are:

- updating webpack and all related package to ones the support v5
- Change dev server command to updated syntax (webpack-dev-server => webpack serve)
- Remove deprecated cli options (-p and --https)
- Change configs settings to new format (node section deprecated and fallback section replaces)
- 'process' polyfill remove from webpack so had to replace with plugin call
- Fixed map file references (which apparently changes without documentation)

Note that using path.resolve was technically not necessary, but making the two config files (test vs. regular) more similar helped in figuring out what one worked and the other didn't (less differences in a diff tool).


package.json
```diff
@@ -16,11 +16,11 @@
     "npm": ">=6 <7"
   },
   "scripts": {
-    "build": "webpack -p --mode production --https false",
-    "build:dev": "webpack --mode development --https false",
-    "build-dev": "webpack --mode development --https false && echo . && echo . && echo . && echo Please use 'build:dev' instead of 'build-dev'.",
+    "build": "webpack --mode production",
+    "build:dev": "webpack --mode development",
+    "build-dev": "webpack --mode development && echo . && echo . && echo . && echo Please use 'build:dev' instead of 'build-dev'.",
```
akrantz, line 21: You could get rid of "build-dev" which was kept for backward-compatibility for a long time now.

test/webpack.config.js
```diff
@@ -5,24 +5,29 @@ const { CleanWebpackPlugin } = require("clean-webpack-plugin");
 const CopyWebpackPlugin = require("copy-webpack-plugin");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
 const path = require("path");
+const webpack = require("webpack");
 
 module.exports = async (env, options) => {
   // const dev = options.mode === "development";
   const config = {
+    target: "web",
```
akrantz, line 13: Please remind me what target does here.

akrantz (commented):
I think it looks OK. It would help to explain more about the changes for webpack v5. Seems like something changed with process, and there is a package added to provide this, and path.resolve() was needed on paths.

package.json
```diff
@@ -16,11 +16,11 @@
     "npm": ">=6 <7"
   },
   "scripts": {
-    "build": "webpack -p --mode production --https false",
-    "build:dev": "webpack --mode development --https false",
-    "build-dev": "webpack --mode development --https false && echo . && echo . && echo . && echo Please use 'build:dev' instead of 'build-dev'.",
+    "build": "webpack --mode production",
```
akrantz, line 19: The `--https false` was intentional to not require dev certs be generated when building for production, so there are not prompts on CI machines. Was there a reason you removed this?

test/webpack.config.js
millerds: https://webpack.js.org/concepts/targets/

I added this just for backwards compatibility good measure because v5 changed the default.

package.json
millerds: Once I upgraded these commands complained about these options not being valid so I removed them.  I could not find an explanation for https not being available anymore.  I did find one that said -p was no longer needs since it was the same a the default or some other flag that went with it (though I can't seem to find that explanation again).

Looks like it passed on the CI machines, so I'm guessing it's OK.

akrantz: It would be nice to retain the ability to turn https on and off. These would be documented under the dev-server options.

test/webpack.config.js
akrantz: Well, it says that 'web' is the default when 'browserlist' is not specified (which we don't use), but still should be OK to not specify this here since the default is still OK.

package.json
akrantz: It seems like -https is still supported: https://webpack.js.org/configuration/dev-server/#devserverhttps

millerds:
The full webpack provided migration document is at https://webpack.js.org/migrate/5/.

The main changes are:
- updating webpack and all related package to ones the support v5 
- Change dev server command to updated syntax (webpack-dev-server => webpack serve)
- Remove deprecated cli options (-p and --https)
- Change configs settings to new format (node section deprecated and fallback section replaces)
- process polyfill remove from webpack so had to replace with plugin call

Note that using path.resolve was technically not necessary, but making the two config files (test vs. regular) more similar helped in figuring out what one worked and the other didn't (less differences in a diff tool).

akrantz (changes requested):
Nice job going through all of these changes. I'm sure this wasn't as easy as you made it look.

--https is not a deprecated option for webpack server, so let's please keep it.

If the path.resolve() isn't needed, wouldn't it be simpler to have it in the previous form. Not a big deal, but I'm not familiar where anything really fundamentally changed here. Did you track down an problem in one of the webpack config files?

millerds:
> Nice job going through all of these changes. I'm sure this wasn't as easy as you made it look.
> 
> --https is not a deprecated option for webpack server, so let's please keep it.
> 
> If the path.resolve() isn't needed, wouldn't it be simpler to have it in the previous form. Not a big deal, but I'm not familiar where anything really fundamentally changed here. Did you track down an problem in one of the webpack config files?

True, I don't need the path.resolve calls now that I've learned more of the difference between the files.  I can revert those (and maybe see if it can be removed from the test config . . . it was already in that one).

I'm not sure what you mean by https is not deprecated . . . leaving those in the cli yields:

[webpack-cli] Error: Unknown option '--https'
[webpack-cli] Run 'webpack --help' to see available commands and options

akrantz:
The --https option now applies when using `webpack serve` but I think what it means is that if using webpack to build without `serve` it is not needed now. Does that mean that it is calling webpack.config.js and has https set to undefined when not using `serve`?

millerds:
The config files have a section that was unchanged an include webpack-dev-server configuration for https (That is the only https config that remains that I know of):

devServer: {
      headers: {
        "Access-Control-Allow-Origin": "*",
      },
      https: options.https !== undefined ? options.https : await devCerts.getHttpsServerOptions(),
      port: process.env.npm_package_config_dev_server_port || 3000,
    },

This stuff will need to get updated when we update webpack-dev-server (next vulnerability).  

I confess that I don't know enough about webpack and it's connection with https configurations.  The commands and the test add-in continue to functions as the did before.  I'm looking to find documentation on the change to remove https cli option, but I'm not finding it all that quickly.

akrantz:
If the --https option specifies ""false", then it won't be undefined, and the values come from office-addin-scripts getHttpServerOptions() function, otherwise, it uses the "false" value.

Because getHttpsServerOptions() can cause a prompt if it needs to generate and install a new cert, we want to ensure this doesn't lead to a prompt when building. You might need to test by using `npx office-addin-dev-certs uninstall` to ensure there isn't an installed cert first, and then trying it.

millerds:
I can give that a try after I get back from my vacation.

There is an artifact setting up the cert in CI machines and the Win tests pass.  However, it looks like Mac had a problem running the tests so maybe that's were we run into problems.

package.json
millerds: that option is for the dev server command which didn't have it before my change and it looks like that is being set in the config file instead of the command line option.

millerds:
I have a mac and set it up to work with this project.  I did the certificate uninstall and when I did the build was asked to install the cert again.  I'm not sure what (if anything) is broken by removing this build option.

igor-ribeiiro (dismissed):

package.json
akrantz: This is to provide the ability to control it via the command line. You shouldn't need to change this. Are you saying there is a reason that you need to do so? Because from what I see this should still work OK.

test/webpack.config.js
akrantz: Could you please check if this works even based on the default if you take this out? I'd rather we only specify things that differ from the default since it is easier to understand that we are deviating from the default for a reason.

test/webpack.config.js
```diff
@@ -5,24 +5,29 @@ const { CleanWebpackPlugin } = require("clean-webpack-plugin");
...
     devtool: "source-map",
     entry: {
       polyfill: ["core-js/stable", "regenerator-runtime/runtime"],
-      test: path.resolve(__dirname, "./src/test-taskpane.ts"),
+      test: "./test/src/test-taskpane.ts",
```
akrantz, line 17: Are you sure there isn't a reason that path.resolve() was being used. What if the command is run specifying a path to the webpack config which isn't the current dir?

akrantz (changes requested):
Please see latest comments.

test/webpack.config.js
millerds: I believe it was your previous comment that suggested that not having the path.resolve(...) was clearer if it wasn't needed.  Since this is in a config file I would presume that using a path to a different config file would not use this setting.

package.json
millerds: As I stated earlier the option is not available in this version of webpack.  The specific output is below (note the error).  The options you point to are for 'webpack serve ...' which is not what we are using to build . . . that case is just 'webpack ...'

PS D:\Repos\Office-Addin-TaskPane> npm run build

> office-addin-taskpane@0.0.1 build D:\Repos\Office-Addin-TaskPane
> webpack --mode production --https false

[webpack-cli] Error: Unknown option '--https'
[webpack-cli] Run 'webpack --help' to see available commands and options

akrantz: There should be `webpack serve` here.

akrantz: OK, maybe I see that since the build doesn't use serve, then the http option doesn't apply for build. But you could do this for "npm run dev-server -- --https false" ?

millerds: - "npm run dev-server -- --https false" leads to "Module not found: Error: Can't resolve 'false' in 'D:\Repos\Office-Addin-TaskPane'"
- npm run dev-server -- --https" doesn't complain, but when I trying and use the addin in Excel says it need a valid signed security certificate.

Note that the our dev-server command didn't use the --https option before (and I believe it did exist in the version of webpack and webpack-dev-server that we were using).

akrantz: OK, the latter is correct. Yes, Excel will require https but it shows that the dev-server is running as http.

akrantz: The dev-server command still supported --https to be used.

akrantz: The build command should now be able to run even without a valid dev cert installed. Do you want to validate that?

```
npx office-addin-dev-certs uninstall
npm run build
```

millerds: The build command does ask for the certificate to be installed if I uninstall it.

akrantz: Yes, this is what I want you to fix. This was the reason for the --https false in the npm build script. There must need to be a different check to see if using "serve" now, so that the https object can just be initialised to false rather than calling to get the dev cert.

millerds: Added and conditional to the config file that allows the build to work without requiring a cert.

test/webpack.config.js
millerds: Removed it.

akrantz: Thanks!

package.json
millerds: removed

akrantz (approved):
Thanks for making the changes.

millerds:
All tests passed . . . not sure why the license/cla check got stuck or how to retrigger that, but it worked before and there shouldn't be anything that changes that.

PR #130, opened 2021-09-08
Update webpack dev server v4

millerds:
Fixes vulnerability


package.json
```diff
@@ -44,6 +44,7 @@
     "@types/node": "^14.14.22",
     "@types/office-js": "^1.0.180",
     "@types/office-runtime": "^1.0.17",
+    "acorn": "^8.5.0",
```
igor-ribeiiro, line 47: When is this getting used?

test/webpack.config.js
```diff
@@ -76,11 +81,13 @@ module.exports = async (env, options) => {
...
     devServer: {
-      contentBase: "testBuild",
+      static: {
+        directory: "testBuild",
+      },
       headers: {
         "Access-Control-Allow-Origin": "*",
       },
-      https: env.WEBPACK_BUILD || options.https !== undefined ? options.https : await devCerts.getHttpsServerOptions(),
+      https: env.WEBPACK_BUILD || options.https !== undefined ? options.https : await getHttpsOptions(),
```
igor-ribeiiro, line 90: Why can't we use the devCerts http function anymore?
Wouldn't it be better for us to update devCerts and import the updated one instead of simply getting rid of the function?

igor-ribeiiro (changes requested):

test/webpack.config.js
akrantz: It is being used. The getHttpsOptions() function is used to map the ca property to cacert since webpack changed the name of the https property. We don't want to change office-addin-dev-certs because this can be used with more than just webpack-dev-server or different versions of it and we don't want a breaking change or to create a different method.

package.json
akrantz: This is a nutty addition.   ;-)

millerds: :-) . . . That was the third commit in the list . . . it's a peer dependency of webpack and we are supposed to be installing peer dependencies.  It should have been added as part of the webpack v5 change, but it was missed and I didn't want to make a separate PR just for this one line.

igor-ribeiiro (approved):

PR #133, opened 2021-09-21
Fix merge problems

millerds:
During merging of template shared changes I found some issues with content being served from webpack, some unnecessary configuration, and some tweaks to make to the new CI/CD pipeline.

### DIFF
--- a/.azure-devops/test.yml
+++ b/.azure-devops/test.yml
@@ -1,9 +1,10 @@
 steps:
-  - script: |
-    echo Running Tests
-    
-    npm run test
-    
-    echo Done running tests
-    
-  displayName: 'Run Tests'
+  - task: CmdLine@2
+    inputs:
+      script: |
+        echo Running Tests
+            
+        npm run test
+            
+        echo Done running tests
+    displayName: 'Run Tests'
--- a/.azure-devops/test.yml
+++ b/.azure-devops/test.yml
@@ -1,8 +1,9 @@
 steps:
-- task: Npm@1
-  displayName: 'npm test'
-  inputs:
-    command: custom
-    verbose: false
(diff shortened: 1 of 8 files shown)


.azure-devops/test.yml
```diff
@@ -1,8 +1,9 @@
 steps:
-- task: Npm@1
-  displayName: 'npm test'
-  inputs:
-    command: custom
-    verbose: false
-    customCommand: 'run test'
-
+  - script: |
```
akrantz, line 2: Isn't it preferrable to use the NPM task? The extra echo statements don't really add anything and you can customize the displayName if desired.

millerds: I found that when using the npm commands we don't always get the command line output and if we do it only happens when the command is done.  The will allow us to see the output as it's happening and more consistently.

Note that the case that are problematic with the npm command are usually the cases that time out (and we can't see the screen that got stuck) and so this won't help as much, but it will show us any command line prompts that cause the test to stop and wait without having to wait an hour.

.azure-devops/test.yml
```diff
@@ -1,8 +1,10 @@
 steps:
-- task: Npm@1
-  displayName: 'npm test'
-  inputs:
-    command: custom
-    verbose: false
-    customCommand: 'run test'
-
+  - task: CmdLine@2
```
igor-ribeiiro, line 2: What does this `@2` mean?
I don't understand the reason some of the commands have `@`. Why is that?

millerds: These are canned actions and the @# is giving you the version of the yaml action being used (kinda like the package.json list of dependencies has a version of the package being used).

igor-ribeiiro (approved):

PR #134, opened 2021-09-24
Add jobs to the pipeline yaml

millerds:
Can't test locally . . . need to try with a PR.


millerds:
Looks like I need get the new file into the master branch before I can test.

PR #135, opened 2021-09-28
Add webview pipeline jobs

millerds:
This will add testing for older webview technologies using the same 1ES win10 image.

### DIFF
--- a/.azure-devops/edgewebview.yml
+++ b/.azure-devops/edgewebview.yml
@@ -1,19 +1,6 @@
 steps:
   - script: |
-     echo Enable EdgeWebView Loopback
-     call npx office-addin-dev-settings appcontainer EdgeWebView --loopback --yes
-     
-     echo Set Edge WebView Registry Settings
-     
-     set PATH1="HKEY_CURRENT_USER\SOFTWARE\Classes\Local Settings\Software\Microsoft\Windows\CurrentVersion\AppContainer\Mappings\S-1-15-2-1310292540-1029022339-4008023048-2190398717-53961996-4257829345-603366646"
-     
-     reg add %PATH1% /f /v DisplayName /t REG_SZ /d "@{Microsoft.Win32WebViewHost_10.0.19041.423_neutral_neutral_cw5n1h2txyewy?ms-resource://Windows.Win32WebViewHost/resources/DisplayName}"
-     reg add %PATH1% /f /v Moniker /t REG_SZ /d "microsoft.win32webviewhost_cw5n1h2txyewy"
-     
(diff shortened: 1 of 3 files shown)


.azure-devops/edgewebview.yml
```diff
@@ -1,19 +1,19 @@
...
-     reg add %PATH2% /f /v ParentMoniker /t REG_SZ /d "microsoft.win32webviewhost_cw5n1h2txyewy"
- 
+      echo Enable EdgeWebView Loopback
+      call npx office-addin-dev-settings appcontainer EdgeWebView --loopback --yes
+      
+      echo Set Edge WebView Registry Settings
+
+      set PATH1="HKEY_CURRENT_USER\SOFTWARE\Classes\Local Settings\Software\Microsoft\Windows\CurrentVersion\AppContainer\Mappings\S-1-15-2-1310292540-1029022339-4008023048-2190398717-53961996-4257829345-603366646"
+
+      reg add %PATH1% /f /v DisplayName /t REG_SZ /d "@{Microsoft.Win32WebViewHost_10.0.19041.423_neutral_neutral_cw5n1h2txyewy?ms-resource://Windows.Win32WebViewHost/resources/DisplayName}"
```
akrantz, line 10: you can use `npx office-addin-dev-settings appcontainer` command to configure this...

millerds: This was copied from the artifact Courtney used to configure out custom VMs before this.  Note the 'npx office-addin-dev-settings appcontainer' command at the beginning.  I experimented with removing the reg key settings and only having the appcontainer command, but the test started failing in the lab with this happened . . . so I added them back (and looks like I adjusted the formatting when doing so).

.azure-devops/full-pipeline.yml
```diff
@@ -14,6 +14,32 @@ jobs:
       - template: ./edgewebview.yml
       - template: ./test.yml
 
+  - job: WebView_EdgeLegacy
```
igor-ribeiiro, line 17: Can we also add the webview to the `Windows_10_Latest` job?

igor-ribeiiro (changes requested):

.azure-devops/full-pipeline.yml
millerds: That job is doing the default case of edge-chromium already.

igor-ribeiiro: I know it is the default case, I just think it would be nice to have it documented as well (the webview for all cases).
I would prefer if it was explicitly said what webview was, just for the sake of avoiding confusion in the future.

millerds: fair enough.

igor-ribeiiro (approved):

PR #136, opened 2021-09-29
Creating unit testing support to taskpane

igor-ribeiiro:
Added import to the newly published `office-addin-mock` and also changed the test structure to accommodate the existence of unit tests now.


package.json
```diff
@@ -28,7 +28,9 @@
     "start:desktop:outlook": "office-addin-debugging start manifest.outlook.xml desktop --app outlook",
     "start:web": "office-addin-debugging start manifest.xml web",
     "stop": "office-addin-debugging stop manifest.xml",
-    "test": "mocha -r ts-node/register test/*.ts",
+    "test": "npm run test-unit && npm run test-end-to-end",
+    "test-end-to-end": "mocha -r ts-node/register test/end-to-end/*.ts",
+    "test-unit": "mocha -r ts-node/register test/unit/*.ts",
```
millerds, line 33: Looks like the convention is 'test:unit' rather then 'test-unit'

package.json
```diff
@@ -28,7 +28,9 @@
     "start:desktop:outlook": "office-addin-debugging start manifest.outlook.xml desktop --app outlook",
     "start:web": "office-addin-debugging start manifest.xml web",
     "stop": "office-addin-debugging stop manifest.xml",
-    "test": "mocha -r ts-node/register test/*.ts",
+    "test": "npm run test-unit && npm run test-end-to-end",
+    "test-end-to-end": "mocha -r ts-node/register test/end-to-end/*.ts",
```
millerds, line 32: I would like something shorter like 'test:hosts' or 'test:integration' or 'test:e2e'

test/unit/getSelectedRangeAddress.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/* global describe, it */
+
+describe(`getSelectedRangeAddress`, function () {
+  it("Using json", async function () {
+    const contextMock = new OfficeMockObject(MockData) as any;
+
+    assert.strictEqual(await getSelectedRangeAddress(contextMock), "C2");
+  });
+});
```
millerds, line 24: This looks like it's testing the mock and not the project.

test/unit/run.ts
```diff
@@ -0,0 +1,35 @@
...
+};
+
+/* global describe, global, it */
+
+describe(`Run`, function () {
+  it("Using json", async function () {
+    const excelMock = new OfficeMockObject(MockData) as any;
+    excelMock.addMockFunction("run", async function (callback) {
+      await callback(excelMock.context);
+    });
```
millerds, line 28: I'm wondering how different this would be from test to test.  If it's going to be standard then maybe we should look at including it in the mock itself.

test/unit/run.ts
```diff
@@ -0,0 +1,35 @@
...
+    global.Excel = excelMock;
+
+    await run();
+
+    assert.strictEqual(excelMock.context.workbook.range.format.fill.color, "yellow");
+  });
+});
```
millerds, line 35: Should probably name the file something more then just 'run'.

test/unit/src/src-file.ts
```diff
@@ -0,0 +1,37 @@
...
+      // Update the fill color
+      range.format.fill.color = "yellow";
+
+      await context.sync();
+      console.log(`The range address was ${range.address}.`);
+    });
+  } catch (error) {
+    console.error(error);
+  }
+}
```
millerds, line 37: Why do we need this?  It looks like a replacement for what the project does.  Tests in the project should be testing project code not independent test code.

test/unit/run.ts
```diff
@@ -0,0 +1,35 @@
+import * as assert from "assert";
+import { OfficeMockObject } from "office-addin-mock";
+import { run } from "./src/src-file";
```
millerds, line 3: Shouldn't we be testing the run code from the project?

millerds (changes requested):

millerds:
I was looking around for a convention for organizing different types of test (e2e vs. integration vs. unit) and didn't find a general consensus.  Not sure if it's better to have top level folders for different types of test vs. sub folders and what the break down of tests should be.

millerds:
Need to consider what get's left when we generate a new project using 'yo office'.  Right now we delete all the test files.  I think at least these unit type tests should be kept.  Without updating the convert script the unit tests will be kept, but we should be intentional about that decision and not accidental.

test/unit/getSelectedRangeAddress.ts
igor-ribeiiro: Agreed, I will remove this unit test from here

test/unit/run.ts
igor-ribeiiro: I guess it can be standard enough that we could add to the mocking package. Basically having a function run on default seems implementable and makes sense. I created this task about it: https://office.visualstudio.com/OC/_workitems/edit/5462138

test/unit/run.ts
igor-ribeiiro: Changing to `excel.ts`, which is the name of the file it is testing.

test/unit/src/src-file.ts
igor-ribeiiro: We don't need it. I was just lazy to fix the project structure to support unit testing. Deleted it.

test/unit/run.ts
igor-ribeiiro: We should and are now :D 
The code the way it was coded, was not very friendly to unit testing with, as it had other stuff on the file, like imports to assets images and Office on ready clauses.
I did some refactoring to make the code more testable.

igor-ribeiiro:
> consensus. Not sure if it's better to have top level folders for different types of test vs. sub folders and what the break down of tests should be.

I like having the top level folders for the different types of tests because they might use totally different technologies and tools.

igor-ribeiiro:
> Need to consider what get's left when we generate a new project using 'yo office'. Right now we delete all the test files. I think at least these unit type tests should be kept. Without updating the convert script the unit tests will be kept, but we should be intentional about that decision and not accidental.

For now, all the tests are deleted, so this would not be shipped to production through the yo office. I think we shouldn't ship this package for yo-office right now, or at least we can do it on other PR without it being an "accident".

src/taskpane/taskpane.ts
```diff
@@ -1,6 +1,39 @@
...
+import * as runOneNote from "./onenote";
+import * as runOutlook from "./outlook";
+import * as runPowerPoint from "./powerpoint";
+import * as runProject from "./project";
+import * as runWord from "./word";
+
+// images references in the manifest
+import "../../assets/icon-16.png";
+import "../../assets/icon-32.png";
+import "../../assets/icon-80.png";
```
millerds, line 11: Can you add 'import "../../assets/icon-64.png";' and 'import "../../assets/icon-128.png";' while your at it.

test/unit/run.ts
millerds: Hmm . . . now that I thing about it some more this is only for our specific run button.  Any dev's code would have different buttons and different calls and this isn't so standard.

test/unit/run.ts
millerds: I don't see how the existing code would create problems with unit testing . . . however the refactoring is nice and make sense.

millerds:
> > Need to consider what get's left when we generate a new project using 'yo office'. Right now we delete all the test files. I think at least these unit type tests should be kept. Without updating the convert script the unit tests will be kept, but we should be intentional about that decision and not accidental.
> 
> For now, all the tests are deleted, so this would not be shipped to production through the yo office. I think we shouldn't ship this package for yo-office right now, or at least we can do it on other PR without it being an "accident".

Makes sense.  Log a task to revisit exporting host specific test (unit and e2e) as part of the project we generate.

millerds:
> > consensus. Not sure if it's better to have top level folders for different types of test vs. sub folders and what the break down of tests should be.
> 
> I like having the top level folders for the different types of tests because they might use totally different technologies and tools.

By top level I'm thinking of siblings of what is not the 'test' folder.  You are using sub-folders of a test folder.  Again, there is not general practice that I've seen so it don't really matter and what's here is OK . . . just wanted to be thoughtful about it.

millerds:
Shouldn't we have unit test cases for the other apps as well?  This would better move us in the direction of eventually making them part of the generated package.

igor-ribeiiro:
> Shouldn't we have unit test cases for the other apps as well? This would better move us in the direction of eventually making them part of the generated package.

Yeah, we should. I will also add a task to that. This will be more of a first usable version for now.

millerds:
> > Shouldn't we have unit test cases for the other apps as well? This would better move us in the direction of eventually making them part of the generated package.
> 
> Yeah, we should. I will also add a task to that. This will be more of a first usable version for now.

Should be simple enough to do right now and avoid having to schedule it in later.

src/taskpane/taskpane.ts
akrantz: You cannot do this in this way. When yo office runs, it makes the project for a single app. This is why there are separate excel.ts etc. files. This is breaking all of that structure.

igor-ribeiiro: @akrantz I was not looking at it before but noticed it later. I just pushed a PR that creates a structure that works with Yo-Office (I just tested here). Could you please have a look?

Now in the generated template from Yo Office: 

- there will be the normal `taskpane.ts` file having the `Office.onReady()` and some import of assets
- `${host}.ts` which will have the `run` function.

millerds: I don't think this is good.  This over complicates the code and results in a projects that has multiple files where only one is need essentially leaving artifact of our repo management in the output instead of a clean project.  Note that your taskpane.ts isn't sharing sharing the code you say it is in the repo . . . it's still repeated across files and we really haven't gained anything.

I get what Adam is reminding us of and we really should just leave these files alone.  If we want to have a common ts file and still have a clean project yo office output we would need the convert to host script to dynamically create the output of the taskpane.ts file . . . and that's not something we are going to do here (if we do it at all).

igor-ribeiiro: Ow, it is just like it is before. The main purpose of I changing those files was to accommodate for unit testing, we already had files with pretty much the same code inside and this didn't change.

I changed the `convertToSingleHost.js` to accommodate my changes. In the end, the customer will only see one more file (the normal `taskpane.ts` file would be split in two).
This needs to be done for unit testing purposes, it is not really feasible to expect to mock a bunch of other code (assets and Office.ready) if you don't care about them.

millerds: Why is it needed for unit testing?  If our framework requires a refactor of simple code like this then we need to recheck our design.

igor-ribeiiro:
> > > Shouldn't we have unit test cases for the other apps as well? This would better move us in the direction of eventually making them part of the generated package.
> > 
> > 
> > Yeah, we should. I will also add a task to that. This will be more of a first usable version for now.
> 
> Should be simple enough to do right now and avoid having to schedule it in later.

Added word and powerpoint tests as well. The rest of the apps don't really have code (other than Project, which we will handle later).

src/taskpane/taskpane.ts
igor-ribeiiro: The assets imports are because we are importing images `.png`. There is really nothing to test on them, and typescript itself doesn't like it. This is not even referent to the unit tests if I import a `.png` the way it is coded typescript complains already. This probably works right now because something of the fact it is being loaded in a HTML makes it work.

The `Office.onReady` is mockable itself. The code inside it is related to HTML and, even if it is possible to mock, I don't think is in the scope of our package, which we are focusing on Office-js code.

So I guess this separation is just of HTML/web-related code and Office-js, which I think is makes sense.

akrantz: The image asserts are there for a reason. If you remove it, it will bring back the issue which led us to add them there.

convertToSingleHost.js
```diff
@@ -33,28 +33,28 @@ async function modifyProjectForSingleHost(host) {
   }
 }
 
-async function convertProjectToSingleHost(host) {
+async function convertProjectToSingleHost(chosenHost) {
```
akrantz, line 36: why is host renamed? unneccessary changes make it difficult to focus on the relevant changes. I suggest renaming this back since it isn't essential here.

convertToSingleHost.js
```diff
@@ -69,8 +69,11 @@ async function convertProjectToSingleHost(host) {
 
   // delete all host-specific files
   hosts.forEach(async function (host) {
+    if (chosenHost !== host) {
+      await unlinkFileAsync(`./src/taskpane/${host}.ts`);
+    }
     await unlinkFileAsync(`./manifest.${host}.xml`);
-    await unlinkFileAsync(`./src/taskpane/${host}.ts`);
+    await unlinkFileAsync(`./src/taskpane/${host}-start.ts`);
```
akrantz, line 76: don't add excel-start.ts or something like that. By trying to alter the basic structure here, it seems to imply that something is wrong with the design...

convertToSingleHost.js
igor-ribeiiro: It was needed because later on this file there is a:
```
// delete all host-specific files
hosts.forEach(async function (host) {
  if (chosenHost !== host) {
    await unlinkFileAsync(`./src/taskpane/${host}.ts`);
  }
  await unlinkFileAsync(`./manifest.${host}.xml`);
  await unlinkFileAsync(`./src/taskpane/${host}-start.ts`);
});
```
Which also names the variable host, so I needed a way to differentiate those two variables.

convertToSingleHost.js
igor-ribeiiro: Well, I'm implying that it can be better, given that currently, we have a bunch of HTML/web-related code within the same file as Office-js code.
I am suggesting separating these two concepts into two files.

src/taskpane/taskpane.ts
igor-ribeiiro: Yeah, I didn't remove them, only changed the file they are in. Now they would be in the `${host}-start.ts` (we can change the name). 

For the template created with Yo Office, the only difference is that now the `taskpane.ts` will only have HTML/web code and the new `${host}.ts` will have the office-js code. Before all the code was in `taskpane.ts`.

millerds: onReady is office.js code . . . so the idea of office.js vs html doesn't line up.  Yes the difference is an extra file in the project output, but as far as the functionality of the add-in goes it doesn't make sense in the project output.  Again . . . we shouldn't have to do this type of refactoring just for the test.

If the test is just importing the run function, why does it matter if the other method or imports are there?  What is the typescript error?

igor-ribeiiro: `onReady` is indeed office.js code. The problem is the code inside it: `document.getElementById` is javascript code that will only work in an HTML.

> If the test is just importing the run function, why does it matter if the other method or imports are there? What is the typescript error?

I don't know why they should matter. I guess typescripts always read the whole file, regardless of the fact that I only imported a specific function. 
By leaving the assets `.png` imports, I get:
```
Office-Addin-TaskPane\assets\icon-16.png:1
�PNG
SyntaxError: Invalid or unexpected token
```

By leaving the `office.onReady` code, I get: 
```
ReferenceError: Office is not defined
```
Because I'm not mocking it.
ReferenceError: Office is not defined

millerds: The e2e test seems to be able to import the run function just fine with the existing files so we should find out why the difference here to make a more informed decision.

We should follow through on our design of the mock . . . if it requires us to also mock Office.onReady then we should do that because that is what out customers will encounter.

Remember . . . this is essentially a validation case for us.  We are having to go through essentially what our customers will when they starting using our mock.  What ever we do is something our customer will have to do as well and we should be careful about what we are asking them to deal with.

igor-ribeiiro: The e2e tests use a manifest and this is the HTML that makes it possible to use the run function.

Yeah, I know this is a validation case. I don't think customers should need to mock `Office.onReady` directly, because they don't need to test our stuff. 
They might need to test the code inside it, which is possible. It just isn't in the scope of the package to test the web-related code, which is what we currently have inside `Office.onReady`, but they can test other code that might be inside `Office.onReady` normally.

millerds: All of the stuff coming out of the template is technically 'our stuff' since the customer hasn't done anything with it.  Once they put their code in there it will be come theirs.  They may not use a run method at all, or maybe they'll change it to try out some things and maybe they'll change the onReady code.  No matter what they do, they'll end up change the test example (once we give it to them) to match their code they want to validate . . . and that might include onReady.  We don't have to validate html change that are happening in the onReady method if we don't want to (though wouldn't it be a simple matter of reading from the document object the values that were updated) or for our purposes you could mock it as empty so the test works and we've mocked everything like customers will end up having to do.

As for the images . . . yeah it looks like the other tests and product code are getting processed by webpack and not directly by typescript (the sideload seems to be OK).  So we should find a typescript way of handling the images.  Again, this is the same thing the customers will have to do.

igor-ribeiiro: Regarding the HTML, I think it still doesn't make sense, as the document element doesn't exist and we would have to mock it.

I am constantly giving examples of clients mocking `Excel.run` and maybe `office.OnReady` but I don't think this is the best practice at all, it was more to give an example that the mocking framework can mock anything.

If we want to give customers the "best" approach, we should separate the code into functions, so they don't even need to Mock those top levels calls. This way the code would be more organized and they wouldn't need to test Office.js code. They shouldn't need to test `Excel.run` (but it is possible) they often need to test what's inside it only.

I will continue with the approach of separating the files in web-related and Office js code. I changed it so now `office.OnReady()` calls a separate function and we only test that function, which is the better practice.

millerds: We aren't trying to test Office.js . . . that isn't the point of any of this.  The point of the mock is so that we can test taskpane code without office.js.  Everything in the project we output is the template code.  If I were writing the add-in from scratch I would make html and code files for my feature and add the unit test code to test the code that I write.  My code could include calling into office with things like Excel.run and Office.onReady along with what ever other logic I need and what ever I write in there would need to be tested.  We need to look at the template code output as if it were all customer code and not ours.

Adding the unit test to our project serves 2 purposes . . . it verifies our mock continues to work with the project, and it will eventually server as an example to out customers.  The example needs to be clean and simple and shouldn't require unnecessary code separation.  We don't need to test onReady, that's not the point, but if it requires a mock in order for it to work, then we can mock that out as an example of what customers will do when the want to override what the production onReady will do.

As for the imports . . . after doing a little research online it looks like a common way to handle this (importing images in typescript isn't unique to us) is to add a custom .d.ts file defining an image module and using a tsconfig file to include it in the typescript build.  These two files would be part of the test code much like there is a test webpack and manifest files to load the test addin code.

If we are still not coming to an understanding it might be better to setup a meeting with you, me, and other interested people to come to a consensus.

convertToSingleHost.js
```diff
@@ -69,8 +69,8 @@ async function convertProjectToSingleHost(host) {
 
   // delete all host-specific files
   hosts.forEach(async function (host) {
-    await unlinkFileAsync(`./manifest.${host}.xml`);
     await unlinkFileAsync(`./src/taskpane/${host}.ts`);
+    await unlinkFileAsync(`./manifest.${host}.xml`);
```
millerds, line 73: Unnecessary change.

src/taskpane/excel.ts
```diff
@@ -18,7 +22,7 @@ Office.onReady((info) => {
   }
 });
 
-export async function run() {
+async function run() {
```
millerds, line 25: Is there a particular reason for this change?  I don't think it's a big deal switching to module.exports . . . but we should be deliberate and purposeful with the changes to avoid unintended side effects.

Same for other files.

millerds (changes requested):

src/taskpane/excel.ts
igor-ribeiiro: Oh yeah, this one is needed. We need to use the `module.exports` to be able to use `require`. And the use of `require` is needed because we need to use it inside a function.

millerds: I don't think that is accurate.  If you have the existing 'export' keyword then it should work just fine.

src/taskpane/office.ts
```diff
@@ -0,0 +1,4 @@
+/* global Office, module */
```
millerds, line 1: I've been thinking about this more and I'm not certain this extra file, this method of mocking, and the need for the extra test framework (jest) is necessary.  You said it was needed for mocking the global 'Office' namespace, but you've already mocked the global 'Excel' namespace without all this extra stuff . . . so why is this different?

test/end-to-end/src/excel-test-taskpane.ts
```diff
@@ -0,0 +1,38 @@
+import { pingTestServer, sendTestResults } from "office-addin-test-helpers";
```
akrantz, line 1: Did this file just get moved, or are there other changes? This shows as an added file rather than a moved and edited file so I cannot tell what's changed.

src/taskpane/excel.ts
```diff
@@ -3,12 +3,16 @@
  * See LICENSE in the project root for license information.
  */
 
+/* global console, document, Excel, module, require */
+
 // images references in the manifest
 import "../../assets/icon-16.png";
 import "../../assets/icon-32.png";
+import "../../assets/icon-64.png";
```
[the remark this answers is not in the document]
millerds: I've created a separate PR for these since I was the one that requested the change.  You are correct that they should be seperate.

src/taskpane/powerpoint.ts
```diff
@@ -27,3 +31,5 @@ export async function run() {
   await Office.context.document.setSelectedDataAsync(" ", options);
   await Office.context.document.setSelectedDataAsync("Hello World!", options);
 }
+
+module.exports = { run };
```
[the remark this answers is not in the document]
igor-ribeiiro: I thought this line was needed since I needed to use `require` in the test (because I need to load inside a function). But turns out I don't need this line.  

I removed it, thanks for noticing that.

src/taskpane/excel.ts
igor-ribeiiro: You are correct. I removed the `module.exports` from these files.

igor-ribeiiro:
> Does convert-to-single-host.js remove the dev dependencies for Jest (and Mocha)?

Yes

src/taskpane/office.ts
igor-ribeiiro: Sure, yeah it is then possible to not need the Office.js file. The only needed thing is to include the `run` file after we already mock Office.
I made some adjustments to reflect that!

src/taskpane/excel.ts
```diff
@@ -4,11 +4,12 @@
...
-import "../../assets/icon-32.png";
-import "../../assets/icon-64.png";
-import "../../assets/icon-80.png";
-import "../../assets/icon-128.png";
+/* eslint-disable no-unused-vars */
+import icon16 from "../../assets/icon-16.png";
+import icon32 from "../../assets/icon-32.png";
+import icon64 from "../../assets/icon-64.png";
+import icon80 from "../../assets/icon-80.png";
+import icon128 from "../../assets/icon-128.png";
```
millerds, line 12: You should add the comment to re enabled the rule after this block.

Same fore other files.

test/unit/powerpoint.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    Text: {},
+  },
+  onReady: async function () {},
+};
+
+// eslint-disable-next-line no-undef
+describe("PowerPoint", function () {
+  it("Run", async function () {
+    const officeMock: OfficeMockObject = new OfficeMockObject(PowerPointMockData);
+    global.Office = officeMock as any;
```
millerds, line 26: Why is this different then the others?  You have a mock for office and a mock for the host in the others, but here you have a single mock that is the office mock (looks like you had the answer to this before you looked into jest).  Is PowerPoint really different in this way?  We should standardize on one way or the other if we can and then add comments explaining the difference if not.  The will help us maintain the code better in the future.

test/unit/excel.test.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const OfficeMockData = {
+  onReady: async function () {},
+};
+
+// eslint-disable-next-line no-undef
+describe("Excel", function () {
+  it("Run", async function () {
+    const excelMock: OfficeMockObject = new OfficeMockObject(ExcelMockData);
```
millerds, line 33: Not a big deal . . . but you are defining the intermediate object in one case (the host) and not the other (Office).  You could skip the intermediate object in both cases and remove the "Data" from the names of the mock object previously define (they seem like more then just data).

It's no big deal and either way can work . . . just looking for consistency.

test/unit/powerpoint.test.ts
igor-ribeiiro: yes, Powerpoint is different, as there is no `Powerpoint` host global var.
What comments do you suggest me putting? I guess the variable's names should indicate what is being mocked in each case.

millerds: Just a simple note saying the objects are based on the API space being using the project (host) code they are testing.

FYI . . . there is a PowerPoint namespace (global var) in the API, it's just not being used in the PP project  which is using the context.document object of the common API (Office namespace/global var).

test/unit/excel.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  run: async function (callback) {
+    await callback(this.context);
+  },
+};
+
+const OfficeMockData = {
+  onReady: async function () {},
+};
+
+// eslint-disable-next-line no-undef
```
millerds, line 30: Why are we doing this instead of adding to the /* global ... */ comment?

PR #137, opened 2021-10-05
Fix ie webview support

millerds:
Office 2019 got broken when upgrading webpack to v5 because it's using the ie webview and webpack is defaulting to a newer coding standard.  The makes webpack use es5 as it's standard so ie webview works.

### DIFF
--- a/.azure-devops/full-pipeline.yml
+++ b/.azure-devops/full-pipeline.yml
@@ -29,18 +29,19 @@ jobs:
         parameters:
           webView: "edge-legacy"
 
-  - job: WebView_IE
-    pool:
-      name: OE-OfficeClientApps
-    steps:
-      - template: ./install.yml
-      - template: ./lint.yml
-      - template: ./build.yml
-      - template: ./devcerts.yml
-      - template: ./edgewebview.yml
-      - template: ./test.yml
-        parameters:
-          webView: "ie"
+  # Need to fix tests (use of find-process module) to be able to support IE 11 webview
+  # - job: WebView_IE
+  #   pool:
+  #     name: OE-OfficeClientApps
+  #   steps:
+  #     - template: ./install.yml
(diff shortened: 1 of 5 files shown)


akrantz (approved):

akrantz:
@millerds It might be nice to consider how to make it easier to retarget for different EcmaScript versions. Perhaps the value specified here could be consolidated into the package.json config along with a comment, to make it easier to change if IE support is not needed.

millerds:
> @millerds It might be nice to consider how to make it easier to retarget for different EcmaScript versions. Perhaps the value specified here could be consolidated into the package.json config along with a comment, to make it easier to change if IE support is not needed.

If I switch to the browerslist way of doing this then that is a package.json change that uses browsers names as the compatibility indicator instead of the ExmaScript version.  I tried it after making the webpack change and it also works.  I think there are more packages that pay attention to this, but I'm not sure what all would be affected.

millerds:
Note that the ie webview version of the tests failed.  It looks like one of the libraries used for our tests (find-process) doesn't work with IE webview.  I confirmed that the project itself (which doesn't have a dependency on find-process) still works, but I'm looking for a way to fix the tests so we can validate the older webviews we need to support.

Given that Office 2019 is currently doesn't work with our templates I might have to check this in with a disabled test to unblock out customers (unless I can find a fix soon).

akrantz (approved):
I think this is OK. But it would be nice to better understand how a developer would change this once IE support (and legacy Edge) is NOT needed. If the add-in targets modern browsers, e.g. the last 2 versions of Chrome, Edge, Safari, Firefox, and this means using Edge WebView2 (Chromium) in Office desktop apps -- or Safari on Mac. This is beyond the scope of this PR, but we should help clarify for users what is required as IE and legacy Edge no longer becomes a requirement.

PR #148, opened 2021-11-01
Fix ie11 test

millerds:
This should this git the test to compile such that the ie11 test can run correctly and be re-enabled.


igor-ribeiiro (approved):

igor-ribeiiro (commented):
Where did you find these find process files?

millerds:
> Where did you find these find process files?

I looked at the code from the find-process package in github and took the portion of them that matched what we needed.  I wanted to narrow it down even more, but that would risk breaking the code that is verified working and would take more time.

tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "allowJs": true,
     "baseUrl": ".",
+    "esModuleInterop": true,
```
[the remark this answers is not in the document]
millerds: Directly for this fix no . . . it does make us more bullet proof for future changes that need to be compatibly with IE webview, but it's not needed for this change directly now that I've added the find code locally (it was part of the first change the turned out not to work).

igor-ribeiiro:
> > Where did you find these find process files?
> 
> I looked at the code from the find-process package in github and took the portion of them that matched what we needed. I wanted to narrow it down even more, but that would risk breaking the code that is verified working and would take more time.

Why isn't the node find-process package working?

millerds:
> > > Where did you find these find process files?
> > 
> > 
> > I looked at the code from the find-process package in github and took the portion of them that matched what we needed. I wanted to narrow it down even more, but that would risk breaking the code that is verified working and would take more time.
> 
> Why isn't the node find-process package working?

I wasn't able to find out specifically why, but it's the use of this library that cause the test in IE webview to stop working . . . the test taskpane fails to load because of a javascript error (I was unable to figure out what specific error).  It's related to trying to compile for down level es5 standard (which we've setup in the compilation process for the taskpane projects).

igor-ribeiiro:
But if you simply copied the code from npm and put it locally, what changed that would make the local code work?

millerds:
It's compiled from our local code rather then just importing the package.  I also tweaked how the main module is defined.  It works for the IE 11 web view (still trying to confirm mac doesn't break).

test/end-to-end/src/find-process-by-name.ts
```diff
@@ -0,0 +1,175 @@
+/* global require, process */
+
+"use strict";
+
+const path = require("path");
+const utils = require("./find-process-utils");
```
akrantz, line 6: You should use `import` and not `require()` with TypeScript.

test/end-to-end/src/find-process-by-name.ts
```diff
@@ -0,0 +1,175 @@
...
+
+function matchName(text, name) {
```
akrantz, line 8: What is all this code doing? I find this hard to read and understand.

test/end-to-end/src/find-process-by-name.ts
```diff
@@ -0,0 +1,175 @@
...
+            cmd: row.CommandLine,
+          }));
+        resolve(list);
+      });
+    });
+  },
+};
+
+export function findProcess(cond): any {
+  const platform = process.platform;
```
akrantz, line 161: Should there be separate functions for each platform? If you're writing your own code to find the process id given a process name, shouldn't this be done separately outside the test code itself? Was this an imported package that we stopped using? Why not have functionality from office-addin-debugging which already has some process stuff in pace?

akrantz (changes requested):
I don't think it's a good idea to write this code in this way. It's not abstracted, organized, readable, or very maintainable. Please see other comments.

millerds:
Managed to find a kill by process name command for the command line I can trigger on the mac.  Much simpler and more direct.

igor-ribeiiro (approved):

millerds:
Discussed offline.

PR #155, opened 2022-03-04
Json project

igor-ribeiiro:
Adding `.json` based options to `convertToSingleHost` script


millerds:
We are not merging this into master . . . it will be its own branch off of master in the primary repo (as opposed to a code fork)

convertToSingleHost.js
```diff
@@ -20,24 +22,24 @@ const readFileAsync = util.promisify(fs.readFile);
 const unlinkFileAsync = util.promisify(fs.unlink);
 const writeFileAsync = util.promisify(fs.writeFile);
 
-async function modifyProjectForSingleHost(host) {
+async function modifyProjectForSingleHost(host, manifestType) {
```
millerds, line 25: Do we need to complicate this with a parameter specifying the manifest type?  I think we start with assuming a json manifest since that is what the template is targeting.

igor-ribeiiro: I was kind of not really just creating a one-off template. This current design could support others hosts in JSON as well.
So I was thinking of just making the more general case first, as it wasn't that hard

igor-ribeiiro:
> 

Do you think it is better? What would be so bad about merging into master?

millerds:
Merging into master would have a down effect of updating and potentially break all of our existing templates.  We need to develop this in isolation, have it tested out and vetted by customers as preview work before trying to combine it with live template code.

convertToSingleHost.js
millerds: The one-off template gives give us a chance to validate and work out the kinks of a json manifest template without requiring more complicated upstream changes (simple new preview template option in generator-office).  Once we are more confident in what the json manifest template would look like as a whole we would have to do a bit more design consideration on making it a single templates template path (and the xml -> json conversion is a simpler way to do that).  

I'm also not sure of the side effects of this.  generator-office would need to be updated and how would older version of gernerator-office handle the new parameter.  This needs to be thought through a little more).

PR #158, opened 2022-03-11
Fixing sideloading on Web not finding manifest file

igor-ribeiiro:
Sideloading in the web was not working because we modified the manifest name to `manifest.dev.xml` and this modification is not expected by `office-addin-dev-settings`, which always looks for the manifest path.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -14,7 +14,6 @@ async function getHttpsOptions() {
 
 module.exports = async (env, options) => {
   const dev = options.mode === "development";
-  const buildType = dev ? "dev" : "prod";
   const config = {
     devtool: "source-map",
     entry: {
@@ -74,7 +73,7 @@ module.exports = async (env, options) => {
           },
           {
             from: "manifest*.xml",
-            to: "[name]." + buildType + "[ext]",
+            to: "[name]" + "[ext]",
             transform(content) {
               if (dev) {
                 return content;


akrantz (changes requested):
We should keep the build type. There are reasons to have separate manifests for prod and dev.

Please fix the problem and not just remove the build type.

millerds:
> We should keep the build type. There are reasons to have separate manifests for prod and dev.
> 
> Please fix the problem and not just remove the build type.

I would assume there are reasons for having separate manifests for prod vs. dev (it makes sense conceptually) and I'm thinking it should be kept, but I'm not aware of what those reasons are for this specifically.  Can you explain?

The problem is that the web pack settings don't match the sideloading script.  The side loading script get its manifest path from the cli argument which refers to its file system location (its used both to read from the manifest and tell sideloading where to get the manifest from).  This works fine for rich client which seems to pull the manifest locally and not from webpack (also doesn't distinguish dev vs. prod).  For the WAC case however, it's using the query parameter sent by sideloading to get the manifest from webpack . . . but webpack has changed the name per webpack config and so the add-in isn't loaded (manifest can't be retrieved).  So the bottom line is that webpack and the sideloading script need to match the name they are looking for and the file has to be readable from the file system.  We're a little baffled as to why this is only being seen now as we haven't changed anything in this area.

There are multiple ways this could be corrected and they each have their own side effects.  Personally, I'm not sold an any one in particular:

- Change webpack to use "manifest.xml" - this ends up removing the option to use dev vs. prod (which right now comes from the same file anyway).  Not sure what we're giving up in this case
- Change the query parameter to assume dev version of the manifest - this would mean that we aren't able to use the prod version anyway as it would always be dev
- Change the file system name of manifest.xml to have both manifest.dev.xml and manifest.prod.xml - Makes it possibly to load either one by changing the script parameter, but then we are having to consider both manifest for the entire project (haven't figured out what that means yet other than package.json scripts are only using one).  This would take a bit more planning and in the meantime the current bug isn't getting fixed.  Note this is somewhat like what TeamsFx (i.e. MetaOS) is doing now.
- Others I'm not considering?

The second and third option experience could be improved by adding a cli parameter to the scripts to indicate dev vs. prod and act accordingly . . . but that is a lot more work to make sure that is plumbed everywhere.  The more I think about this, the more I suspect the dev vs. prod manifest support isn't really there and should be a feature we fully develop.  I think the dev vs. prod in webpack should be kept, but I have yet to discover a better alternative when the current behavior is broken.

igor-ribeiiro:
Thanks, @millerds for listing some possible options. I agree that we currently don't have an easy way to distinguish between prod and dev, as we can only make those arguments make sense if we are going to have two separate manifests.
What do you think of Darren's option @akrantz?

As for the duration of the error, I can suspect it might be happening for months and we just didn't notice. The tutorial explicitly said that if the sideload doesn't work, just use manual loading. So my guess is that a lot of customers just manual sideloaded and this error existed for months now.

akrantz:
I'd like to understand better how this used to work before. We already had the manifest in the build output (dist folder) be named with the build type (manifest.dev.xml / manifest.prod.xml), and the sideload to web should have known to use manifest.dev.xml. Now, there could be confusion between the dist folder containing manifest.dev.xml and the root folder containing manifest.xml. With webpack-dev-server it could be serving up only the dist folder contents, or it might be that before the change to webpack 5, it was serving manifest.xml as well.

When we develop the templates, we only really think about the "dev" experience, and not about what's needed to work with "production" bits. There needs to be a different SourceLocation for Production, which requires a different manifest. The goal was to have a single manifest.xml which is then customized in the build to produce dev/prod specific manifests with the appropriate values. This would allow for example, debugging against prod bits. 

If the issue is that office-addin-dev-settings sideload needs to know whether it is using dev or prod, it shouldn't be difficult. First, it could look for manifest.dev.xml, then manifest.prod.xml, then manifest.xml and use the first one that is available. For a particular build, there should not be both dev/prod in the same dist folder.

I would also prefer to revert the webpack config change so the file is named `manifest.dev.xml`. I'm not sure what the current name is now, but the use of two periods was intentional when I did it.

millerds:
So are you suggesting that office-addin-dev-settings ping webpack to find one of a set of possible manifest files and then use that for the query parameter?  An 'npm build' command creates the dist folder with manifest.prod.xml, but when you run 'npm start' it will serve up manifest.dev.xml regardless of what is in the dist folder so we can't depend on that.

If I'm using win32 clients it's not reading it from webpack or the dist folder.  As far as I can tell it's reading straight from the manifest.xml location in the project and not pulling from web pack.

I get the desire to have prod vs. dev debugging etc. but it looks like what we are providing the customers right now is partially that.  The manifest content is the same either way . . . the file is just being named dev vs. prod in webpack to they are always getting the same thing even if we were to retrieve a file that was named one or the other.  More work to provide the full scenario is needed, but the web debugging experience is broken now.

akrantz:
No, a build is done for both "npm build" (prod) and "npm start" (dev). For prod builds, it writes to disk in the dist folder. For dev builds, it serves from memory but it is still a dist folder.  I think the key thing here is that when dev is getting it from `https://localhost:3000/` then querying `https:/localhost:3000/manifest.dev.xml` should be the manifest.dev.xml which is in the in-memory build output.

It is possible to have dev builds also write to disk, but there is usually no need.

For Excel-Custom-Functions, there were some hacks to get around the fact that the dev-server in-memory dist folder doesn't work for the custom functions runtime since it isn't a browser context. So it uses the on-disk stuff which is then also served from the in memory, and this can be causing confusion about the simpler thing needed for non-custom-functions templates. I can explain more on a call if needed.

igor-ribeiiro:
> 

I might be missing something. If the `dist` folder has only one: the `.dev` or the `.prod` manifest at a time and you are suggesting we just look for a file with one of the two extensions, what is the difference of simply removing those extensions all together? 
The customer can still build for production with no problem. We are just removing the `.ext` in the filename. The behavior for dev and prod webpack will still be the same

akrantz:
It's difficult for me to remember. First, there were scenarios where the root folder could have multiple manifests, e.g. dev and prod, but could also be for different rings. While the default name is manifest.xml, it should be possible to use a different name if specified in the argument. Second, there might have been something related to Excel-Custom-Functions, but it's hard to remember.

I'd like to first understand why it's not working now. Was it just the manifest file name change? Or something else related to webpack 5?

None of this should require a particular manifest file name, so it's good to understand the details first before speculating on changes needed.

igor-ribeiiro:
> 
First question:
It is possible to use different manifest names. There is no hardcoded manifest name. So no bug in there.

Second:
I just tested custom functions on the web and is not even sideloading right now and it appears to be a different bug. I will create a bug for us to track why it is not working for web.

"I'd like to first understand why it's not working now. Was it just the manifest file name change? Or something else related to webpack 5?"
I think my code change might have not fixed the bug. I started getting it again today. I will do some more investigation to make sure, but maybe this problem is unrelated to the manifest name

igor-ribeiiro:
Apparently, I was doing these final tests wrong and this PR indeed fixes the bug (it is sideloading as expected).
As previously mentioned, Office-Addin-Scripts is not expecting a change in the manifest name, which is happening in the webpack.
I'm with Darren that we don't have a user scenario that needs dev and prod in the way implemented now.

For excel-CFs, I will try more later, but I was having troubles sideloading on the web (might be the bug Courtney is solving for excel). 
Searching in the docs, we don't mention anywhere about this prod vs dev difference, so I don't think customers are getting advantage of this feature right now.

Any comments @akrantz?

akrantz:
I thought you said that the problem still occurs even with the manifest not including dev/prod in the name.

I still would like to know WHY the sideload is failing with the different name. It seems like the bug may really be in the sideload code rather than this. The manifest name is either explicitly provided to sideload, or it happens implicitly. Obviously if passed explicitly it would work, so this implies that the code for sideload to web which is composing the url to read the manifest from the dist folder served from memory by webpack-dev-server should include the build type as well, e.g. `https://localhost:3000/manifest.dev.xml`.

igor-ribeiiro:
> I thought you said that the problem still occurs even with the manifest, not including dev/prod in the name.

It was an error on my part. The problem does not occur when removing prod/dev.

We know why the problem happens. We can fix on the sideload side to add the build type or remove build types. Both approaches should work.
By changing the sideload code, what I could do is make it check for `manifest-path` normally, if it doesn't exist check for `manifest.dev`, else `manifest.prod`.

I agree with Darren that we don't currently provide much value with the `prod` vs `dev` and this makes the experience more confusing.
Do you think there is much value in maintaining this prod vs dev? I personally think it is just complicated our templates and we take this bug to clean up these.

millerds:
Your project doesn't even include the tooling use to sideload and the manifest files are named with 'prod' in the project itself which is not how the templates are setup.  There is certainly value in being able to have different manifests for prod vs. dev, but I don't see that in the templates as they are.  A customer would have to create different manifests (we only provide one) and update package.json scripts (or use extra command line arguments) to use those different manifests and possibly update webpack themselves depending on the name of the manifests they create.  The only way that current webpack works is if they create the manifest with the right prod vs. dev name that we aren't giving them.

igor-ribeiiro:
@akrantz do you think you can be fine with this change? It would make our templates simpler as a benefit for it

millerds (approved):

Issue #156, opened 2022-03-08
unneeded settings in tsconfig.json and package.json


Rick-Kirkham:
UPDATE 1/9/24: The webpack.config.js change was made in a recent PR. Only the changes to tsconfig.json and package.json still need to be done.

# Prerequisites

Please answer the following questions before submitting an issue. 
**YOU MAY DELETE THE PREREQUISITES SECTION.**
- [x ] I am running the latest version of Node and the tools
- [x ] I checked the documentation and found no answer
- [x ] I checked to make sure that this issue has not already been filed

# Expected behavior

webpack.config and tsconfig.json does not have unnecessary/confusing settings

# Current behavior

The following "rule" is in the webpack.config for this template. This would be relevant for the React version, but not for this version. Having this unneeded loader may confuse devs (like me). 
```
        {
          test: /\.tsx?$/,
          exclude: /node_modules/,
          use: "ts-loader",
        },
```

The tsconfig.json file has a "jsx" property, with value "react". This "jsx" property should be deleted, because this is not a React project.

Finally, ts-loader should be removed from the package.json. It is not actually doing anything. 

I have tested all these changes and the project works fine.

akrantz:
This is needed for TypeScript, regardless of whether using React.
This is not present in https://github.com/OfficeDev/Office-Addin-TaskPane-JS.

akrantz:
Because both .ts and .tsx are valid extensions for TypeScript files. We don't include any .tsx files by default, but users can add them.

Both loaders are needed. If I remember correctly, the ts-loader handles compiling TypeScript to JavaScript, and Babel handles the polyfills.

Rick-Kirkham:
Sorry, but this is not consistent with my understanding of webpack.config. Polyfills are handled by plug-ins, which is a later section of the webpack.config, not by loaders. This babel loader is configured to transpile `ts` files to JavaScript. 

Yes it is possible that users will convert their add-in to React and add `tsx` files, but they could change the project in lots of ways that would require changing the webpack.config settings. This is just one. It is confusing to have this in a project that doesn't need it.

akrantz:
Ricky, it doesn't hurt anything to keep it like this, but changing does break stuff. Can you please show an example of a React and non-React TypeScript project using a modified webpack config? Because I know specifically that this needs to be this way based on implementing it before.

Rick-Kirkham:
[...]

PR #159, opened 2022-03-14
Remove confusing React stuff from this non-React project template

Rick-Kirkham:
I've tested this and the project works fine with these changes. This will make the project less confusing to developers who are poking around in the config files, which they might be doing either for learning or to modify the project. 

The only change I intentionally made to package.json is to delete this line:

"ts-loader": "^9.2.5",

But I see a lot of other deletions in the diff. Maybe one of you guys can explain that.

This will fix [156](https://github.com/OfficeDev/Office-Addin-TaskPane/issues/156)


akrantz (changes requested):
You have proposed a lot of changes that are not in line with how these templates work. For example, don't remove convert-to-single-host, as this breaks yo office.  All of the other changes to the engines, scripts are also necessary. All of the dependencies you are removing are necessary.

Furthermore, this template is merged into Office-Addin-TaskPane-React and as I mentioned before, unless there is actually a problem shown by having the .tsx extension, we should not be making these changes.

Rick-Kirkham:
> You have proposed a lot of changes that are not in line with how these templates work. For example, don't remove convert-to-single-host, as this breaks yo office. All of the other changes to the engines, scripts are also necessary. All of the dependencies you are removing are necessary.
> 
> Furthermore, this template is merged into Office-Addin-TaskPane-React and as I mentioned before, unless there is actually a problem shown by having the .tsx extension, we should not be making these changes.

Thanks for the feedback @akrantz. You actually responded before I could finish creating the PR comment.  
I don't understand all the deletions in the package.json, so I need some help with this. E.g., I didn't intend to remove convert-to-single-host. I'll see if I can fix this in another commit to the PR. 

If this template is merged into Office-Addin-TaskPane-React, then I think we should change this architecture so that the React-related things are only in the Office-Addin-TaskPane-React template. 

As I suggested in the Issue 156, there is a problem with the having React-related things in a non-React template: It is confusing to developers.

akrantz:
We don't have React things in the non-React template. We just have the extension for both .ts and .tsx files. Is it just the file extension filter? We've never gotten feedback that this is confusing (other than for you). 

The ts-loader I believe is for type checking. Did you test out having errors in the TypeScript code so they are caught by the type checking? There are complex aspects to this and I don't even recall all of it. Unless there was a change with webpack 5 which eliminated the need for ts-loader, I'm pretty sure we should still include it. I could not tell easily from the webpack docs.

Rick-Kirkham:
> We don't have React things in the non-React template. We just have the extension for both .ts and .tsx files. Is it just the file extension filter? We've never gotten feedback that this is confusing (other than for you).
> 
> The ts-loader I believe is for type checking. Did you test out having errors in the TypeScript code so they are caught by the type checking? There are complex aspects to this and I don't even recall all of it. Unless there was a change with webpack 5 which eliminated the need for ts-loader, I'm pretty sure we should still include it. I could not tell easily from the webpack docs.

I'll try the type-checking. Regardless of the result, I will delete this PR. I think it makes more sense for me to just raise the issue and leave implementing the fix to the developers.  

Being the customer's advocate is part of my job. If I find something confusing, there is a good chance that customers will too.

PR #164, opened 2022-04-06
Json cleanup

igor-ribeiiro:
Updating other .xml references in the JSON-based template.

### DIFF
--- a/.azure-devops/test.yml
+++ b/.azure-devops/test.yml
@@ -9,8 +9,8 @@ steps:
       script: |
         
         echo Setting WebView Type: ${{ parameters.webView }}
-        call npx office-addin-dev-settings webview manifest.xml ${{ parameters.webView }}
-        call npx office-addin-dev-settings webview test/end-to-end/test-manifest.xml ${{ parameters.webView }}
+        call npx office-addin-dev-settings webview manifest.json ${{ parameters.webView }}
+        call npx office-addin-dev-settings webview test/end-to-end/test-manifest.json ${{ parameters.webView }}
         
         echo Running Tests
         npm run test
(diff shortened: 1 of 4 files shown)


.azure-devops/test.yml
```diff
@@ -9,8 +9,8 @@ steps:
       script: |
         
         echo Setting WebView Type: ${{ parameters.webView }}
-        call npx office-addin-dev-settings webview manifest.xml ${{ parameters.webView }}
-        call npx office-addin-dev-settings webview test/end-to-end/test-manifest.xml ${{ parameters.webView }}
+        call npx office-addin-dev-settings webview manifest.json ${{ parameters.webView }}
+        call npx office-addin-dev-settings webview test/end-to-end/test-manifest.json ${{ parameters.webView }}
```
millerds, line 13: Did you walk through the code to figure out what this would do?  This may be something we've missed in our planning.

igor-ribeiiro: This part of the code is related to the pipelines, so we need to change it in order for them to have the correct manifest references.
These commands don't seem to do anything special (mostly setting reg keys). So it looks like we can keep them the same for build

millerds: Yes . . . the yaml files are for the pipelines . . . but what effect does the command have?  I think you are right in that it sets a reg key in the dev settings, but how is the manifest used to create that key?   Does the key have some reference to the json manifest and if so how is that used later?  Since it has to do with the host loading the add-in I want to make sure the host isn't getting mixed up as a result (which will have to be fully verified later).

igor-ribeiiro: The only part of the manifest that gets used is the ID. So the reg key creates only sets the ID and has no further information about the type of the project.
So I think we don't need to worry about host problems in other parts of the code

millerds (approved):

PR #170, opened 2022-05-06
Add postinstall .Net check

millerds:
This will indicate weather or not .Net is installed (needed by the template) at the end of running 'npm install'

### DIFF
--- a/checkDotNet.js
+++ b/checkDotNet.js
@@ -1,0 +1,22 @@
+const chalk = require("chalk");
+const childProcess = require("child_process");
+
+/* global console require */
+
+try {
+  // Find the .Net runtimes installed
+  let result = childProcess.execSync("dotnet --list-runtimes");
+  const pattern = /(?<=Microsoft.NETCore.App )[\d.]+/g;
+  const matches = result.toString("utf-8").match(pattern);
+  let foundDotNet = false;
+
+  // Look for version 5 or greater
+  matches?.forEach((match) => {
+    const major = parseInt(match.split(".")[0]);
+    foundDotNet = foundDotNet || major >= 5;
+  });
+
+  console.log("Correct version of .Net is installed");
+} catch (err) {
+  console.log(chalk.bold.red(".Net 5 or greater is required for json manifests."));
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "office-addin-taskpane",
       "version": "0.0.1",
+      "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
         "core-js": "^3.9.1",
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "dev-server": "webpack serve --mode development",
     "lint": "office-addin-lint check",
     "lint:fix": "office-addin-lint fix",
+    "postinstall": "node checkDotNet.js",
     "prettier": "office-addin-lint prettier",
     "start": "office-addin-debugging start manifest/manifest.json",
     "start:desktop": "office-addin-debugging start manifest/manifest.json desktop",


akrantz:
Shouldn't this be provided from office-addin-scripts instead of directly in office-addin-taskpane?

checkDotNet.js
```diff
@@ -0,0 +1,22 @@
...
+
+  // Look for version 5 or greater
+  matches?.forEach((match) => {
+    const major = parseInt(match.split(".")[0]);
+    foundDotNet = foundDotNet || major >= 5;
+  });
+
+  console.log("Correct version of .Net is installed");
+} catch (err) {
+  console.log(chalk.bold.red(".Net 5 or greater is required for json manifests."));
```
igor-ribeiiro, line 21: Given that we are catching this error and letting the process continue, I think this message should also include something telling the customer to install `.net` 5.
Like:
`Please install the latest .net 5 or greater`

igor-ribeiiro (changes requested):

igor-ribeiiro:
> Shouldn't this be provided from office-addin-scripts instead of directly in office-addin-taskpane?

It is also in the scripts when going to sideload.
We have thought that having in both places could warn customer quicker about this requirement

akrantz:
It's going to be a nightmare to maintain in the project itself and merge across. You're better off having it only in the office-addin-scripts package.

akrantz:
You should just have whatever command is needed to run give an error if .NET isn't installed when it needs to be. This should not be done in the project itself.

millerds:
This is specific to the json manifest template and only an intermediate step for build.  

.Net will be needed when a developer tries to sideload the json manifest project.  I've already added the check to sideloading in office-addin-dev-settings (behind the scenes) to say .Net needs to be installed.  Louis was suggesting that we need to advise devs sooner that this is a requirement and not depend on documentation.  I thought about exporting the method in office-addin-dev-settings and using it, but we will be taking this back out in just a couple more months (when more work in outlook and the new MetaOS service are done) and I wondered if we wouldn't run into some longer-term maintenance costs if it were publicly supported in the scripts package.  The risk is probably small and if we were intending to support this longer term I would definitely do it via the scripts package.  This seems like it would be easier to remove later on.

akrantz:
If the check is in office-addin-dev-settings, I don't see why it needs to be here.

millerds:
> If the check is in office-addin-dev-settings, I don't see why it needs to be here.

Personally I could go either way (both or just sideload) . . . the point brought up comes down to user experience.  They create a new project, npm install, but then when they do npm start there a new error saying they need .Net 5 runtime and we aren't installing it for them (too complicated to do via node/java).  Is that too annoying for developers not to have the message earlier that they need an extra manual install?

akrantz:
Sure, but the javascript code to detect .NET is missing is already in office-addin-dev-settings. Why go and add script in the project itself as well? It's simpler and better to have the checks where needed.

akrantz:
You could still make the postinstall step run a command in office-addin-scripts to achieve the same thing.

postinstall: office-addin-dev-settings check-dotnet-installed

millerds:
It boils down to 3 options:

1. Only have the check in sideload. 
Pro: only one place in code to deal with at the point it's needed.  
Con: developers aren't notified of the dependency until later in the loop
2. Check in sideload and separately check on package install.  
Pro: developers notified sooner of the dependency and no public method.  
Con: similar code in two places
3. Check in sideload and call check on package install. 
Pro: developers notified sooner and code is shared.  
Con: Potential for new dependency on method that will be deleted in a short time.

It's a tradeoff no matter which way we go.

akrantz:
No, you can have the code only in the package, but still provide the check in multiple places. Nothing here requires the code to go in the project itself.

akrantz:
I think this is your step 3, but I don't see what the con is. There has been and will continue to be a dependency on office-addin-dev-settings regardless of whether or not this check is done. If you're saying it needs to be a direct dependency on office-addin-debugging, then I would say export the functionality from there so that package can be called.

millerds:
> I think this is your step 3, but I don't see what the con is. There has been and will continue to be a dependency on office-addin-dev-settings regardless of whether or not this check is done. If you're saying it needs to be a direct dependency on office-addin-debugging, then I would say export the functionality from there so that package can be called.

The Con for option 3 is that in order to use it method in the project it has to be exported from the package.  This would also make it available for anyone to import our package and use in their own project in their own way.  We would then take it away just a couple months later.  This is a small window, but still a possibility.

millerds:
I occurs to me that another con is potentially breaking projects generated with 'yo office' when we remove the method.  Developers who update the their packages after we remove the backward conversion would still have the postinsall script pointing to it that would need to be cleaned up by them.  If the check is local the project would continue to work after we remove it, it would just be checking for .Net when it didn't need to anymore.

akrantz:
There is not a big cost to add it to the package, nor remove it later. The cost is how you manage the script and changes to it and merge it across the templates, and then when you remove it, customers which have a certain version have to remove it manually rather than having the package updated to say this is no longer supported and tells them to remove the script line.

millerds:
This is only ever going to be provided to our json manifest template and not merged to the others.  We haven't finalized plans for json manifest templates (how many, how are they managed, etc), but this will be removed before we have multiple json manifest based templates.

PR #176, opened 2022-06-13
Wait for host before killing process

millerds:

---

**Change Description**:
When Excel tests complete we need to close the workbook before the host, but we kill the process to quickly.  We need to wait for Excel (and other hosts) before killing the process.

    No.

    No.

    No.

    No.

    Ran automated tests multiple times.

### DIFF
--- a/test/end-to-end/src/test-helpers.ts
+++ b/test/end-to-end/src/test-helpers.ts
@@ -27,6 +27,7 @@ export async function closeDesktopApplication(application: string): Promise<bool
       throw new Error(`${application} is not a valid Office desktop application.`);
   }
 
+  await sleep(3000); // wait for host to settle
   try {
     let cmdLine: string;
     if (process.platform == "win32") {


igor-ribeiiro (approved):

test/end-to-end/src/test-helpers.ts
```diff
@@ -27,6 +27,7 @@ export async function closeDesktopApplication(application: string): Promise<bool
       throw new Error(`${application} is not a valid Office desktop application.`);
   }
 
+  await sleep(3000); // wait for host to settle
```
akrantz, line 30: Why is this necessary?

millerds: In Excel specifically we were at one point calling close workbook, but not closing Excel and this would leave the app open after tests.  It was changed to close Excel after calling close workbook, but it seems that the closing of Excel is happening too quickly after calling close workbook and the next time Excel is opened the recovery pane (or on mac a dialog) is opened.  By putting the pause here we are allowing any final host operations (like close workbook) to happen before killing the process and avoiding any recovery states.

PR #182, opened 2022-09-12
VS Code launch.json no longer needs useWebView

akrantz:
From https://github.com/microsoft/vscode/issues/157587 I've learned that the useWebView is no longer needed for debugging with the Edge webview.

I validated that debugging an Excel task pane works even with this property removed.

No.

Yes

No.

I don't think so.

1. Opened the Office-Addin-TaskPane project in VS Code.
2. Ran the dev server.
3. Opened file excel.ts and set a breakpoint on line 25 which contains: range.load("address");
4. From Debug tab, selected "Excel Desktop (Edge Chromium)"
5. Click the play button to start debugging. 
6. When Excel opened, I clicked the ribbon button to open the task pane for the add-in, and clicked the Run button.
7. Hit the breakpoint.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,9 +21,7 @@
         "${workspaceFolder}/test/end-to-end/*.ts"
       ],
       "internalConsoleOptions": "openOnSessionStart",
-      "runtimeArgs": [
-        "--preserve-symlinks"
-      ]
+      "runtimeArgs": ["--preserve-symlinks"]
     },
     {
       "name": "Debug Unit Tests",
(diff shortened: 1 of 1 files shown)


millerds (approved):
Thanks Adam!  I have been noticing this myself and couldn't find good documentation on it either, but hadn't gone so far as to log an issue for it.

igor-ribeiiro (approved):

.vscode/launch.json
```diff
@@ -21,9 +21,7 @@
         "${workspaceFolder}/test/end-to-end/*.ts"
       ],
       "internalConsoleOptions": "openOnSessionStart",
-      "runtimeArgs": [
```
akrantz, line 24: I just noticed there is a diff here. Is it desired to keep the existing formatting?

millerds: I'm not worried about it give the single array element.

PR #192, opened 2022-12-26
Fix webpack html hot-update

QuanfuXiao:


millerds (dismissed):

millerds:
Looks like the earlier failure was an odd repeated fluke . . . especially since I noticed the test webpack config wasn't part of the change.  You should update the test webpack.config.js and we'll run the tests again.  It should pass and we can check in.

PR #200, opened 2023-01-26
Remove browser launch and task json

millerds:

---

**Change Description**:
Current functionality of the sideloading cli doesn't work with vs code launch config options.  Sideloading needs to be updated in order to support it.  Removing for the time being.

Nol

Yes.  Removing some options that don't work.

Yes.  VS code support files have been updated.

Possibly . . . if there is anything the specifically refers to debugging office online then it will need to be looked at.

    Describe manual testing done.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -125,21 +125,6 @@
       "webRoot": "${workspaceRoot}",
       "preLaunchTask": "Debug: Word Desktop",
       "postDebugTask": "Stop Debug"
-    },
-    {
-      "name": "Office Online (Chrome)",
-      "type": "chrome",
-      "request": "launch",
-      "webRoot": "${workspaceFolder}",
-      "preLaunchTask": "Debug: Web"
-    },
-    {
-      "name": "Office Online (Edge Chromium)",
-      "type": "msedge",
-      "request": "launch",
-      "port": 9222,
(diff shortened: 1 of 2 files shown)


Rick-Kirkham:
@millerds Is this the only template that is affected? Do we have these same launch/tasks in other templates?

millerds:
> @millerds Is this the only template that is affected? Do we have these same launch/tasks in other templates?

This will get merged into the other templates as well.

Rick-Kirkham (approved):

Rick-Kirkham:
> > @millerds Is this the only template that is affected? Do we have these same launch/tasks in other templates?
> 
> This will get merged into the other templates as well.

Got it. There is actually only one section of one article in the docs which uses this option. I'll remove that from the article.

akrantz:
What's the rationale for removing this?

millerds:
We talked in person . . . but essentially there is a conflict between the launch config needing the URL and the office-addin-debugging web cli needing the document URL and both trying to launch the browser.  My investigation shows that it's been broken for a while and we haven't seen complaints so I'm removing the launch config to give us time to come up with a solution to the conflict.

millerds:
Can I get and OP DevX review?

Adrian-MSFT (approved):

kwame-ansong (approved):

PR #220, opened 2023-11-23
Draft PR: default-ts template for WXP on Toolkit

YueLi-MSFT:

---

**Change Description**:

    Describe what the PR is about. 

    If Yes, briefly describe what is impacted.

    If Yes, briefly describe what is impacted.

    If Yes, briefly describe what is impacted.

    If Yes, briefly describe what is impacted.

    Describe manual testing done.


.vscode/launch.json
```diff
@@ -25,5 +25,68 @@
       "preLaunchTask": "Debug: Outlook Desktop",
       "postDebugTask": "Stop Debug"
     },
+    {
```
millerds, line 28: Should probably order the configs in the same ways as is done in the master branch to maintain uniformity.

package.json
```diff
@@ -7,7 +7,6 @@
   },
   "license": "MIT",
   "config": {
-    "app_to_debug": "outlook",
```
millerds, line 10: This should not be removed for this special case.  If it doesn't look like it's needed then we should log a bug to remove it in general across all the templates.

millerds (changes requested):
Note that the current json manifest branches were reduced from the master branch to target Outlook because that is all that was supported at the time.  When changing it back to support other host we should continue to follow the patterns used in general and if there is a desire to change the patterns then a task/bug should logged and discussed as a change for all templates.

We should be following the branch progression for changes . . . meaning changes for json manifest template goes to "json-preview" branch first.  Changes being made ready for public consumption should then be merged from "json-preview" to "json-preview-yo-office-prerelease" and verified using 'yo office' there before finally being pushed to "json-preview-yo-office" for customers to use.

On top of following the branch progression, we need to coordinate with the documentation team (Ricky) and managing the branches and Outlook template (which is going to GA in a couple of months) as they own the repo management now.

package.json
```diff
@@ -25,7 +24,10 @@
     "signin": "office-addin-dev-settings m365-account login",
     "signout": "office-addin-dev-settings m365-account logout",
     "prettier": "office-addin-lint prettier",
-    "start": "office-addin-debugging start manifest.json",
+    "start:word": "office-addin-debugging start manifest.json --app word",
+    "start:excel": "office-addin-debugging start manifest.json --app excel",
+    "start:powerpoint": "office-addin-debugging start manifest.json --app powerpoint",
+    "start:outlook": "office-addin-debugging start manifest.json --app outlook",
```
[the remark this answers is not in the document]
QuanfuXiao: Thanks Darren, it is reasonable to leave one command in "package.json" and pass parameters in tasks config.

.vscode/launch.json
QuanfuXiao: @millerds , should we use unique port for each config?

millerds: No . . . the ports listed are the standard cases for attaching to the launched webview host.  The chrome version and the legacy version each have that as their default.

I'm just asking that the order of each config match that of the master branch.

package.json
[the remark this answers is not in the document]
millerds: That is what the master branch does . . . use that as your starting point.

.vscode/tasks.json
```diff
@@ -34,10 +34,52 @@
         "showReuseMessage": false
       }
     },
+    {
```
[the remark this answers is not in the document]
YueLi-MSFT: Thanks Millerds! Will do!

.vscode/tasks.json
```diff
@@ -69,6 +111,20 @@
       },
       "problemMatcher": []
     },
+    {
+      "label": "Link",
+      "type": "shell",
+      "command": "npm link office-addin-dev-settings --force",
```
millerds, line 117: I don't think we want to do this automatically as part running in VS Code.  This should be left for someone to use if they explicitly want to try a change to the scripts package locally.

QuanfuXiao: Yeah, it is only for POC demo and will be deleted when TAOS sideload for WXP manifest is ready.

QuanfuXiao:
Hi @millerds , regarding the branch issue, is it reasonable to add a new branch only for the toolkit? Since "json-preview-yo-office" should remain to be used in Yo Office for Outlook only.

millerds:
> Hi @millerds , regarding the branch issue, is it reasonable to add a new branch only for the toolkit? Since "json-preview-yo-office" should remain to be used in Yo Office for Outlook only.

This is something that needs to be discussed with the docs team that jointly owns maintaining the templates.  I plan on figuring that out and we'll have to have a plan in place before modifying the existing branches.

.vscode/launch.json
YueLi-MSFT: Got it. Thanks @millerds

.vscode/tasks.json
```diff
@@ -60,8 +102,22 @@
     },
     {
       "label": "Install",
-      "type": "npm",
-      "script": "install",
+      "type": "shell",
```
millerds, line 105: why is this needed.

manifest.json
```diff
@@ -184,7 +184,7 @@
                     {
                       "id": "msgReadOpenPaneButton",
                       "type": "button",
-                      "label": "Show Taskpane",
+                      "label": "Hello Excel",
```
millerds, line 187: We should not be labeling this specific to one host.  This is used by all the hosts.

src/commands/commands.ts
```diff
@@ -13,16 +13,17 @@ Office.onReady(() => {
...
-  const message: Office.NotificationMessageDetails = {
-    type: Office.MailboxEnums.ItemNotificationMessageType.InformationalMessage,
-    message: "Performed action.",
-    icon: "Icon.80x80",
-    persistent: true,
-  };
-
-  // Show a notification message
-  Office.context.mailbox.item.notificationMessages.replaceAsync("ActionPerformanceNotification", message);
+async function action(event: Office.AddinCommands.Event) {
```
millerds, line 16: This template is used for Outlook too.  If you want to make a command sample for different hosts (something not done in current xml based templates) then you need to make the template handle all hosts, or make options for host specific stuff that is cleaned up by the convert script when instantiating template for a particular host.

.vscode/tasks.json
QuanfuXiao: Will change it back. We added a "npm link" script for demo, but it seems not work well with "npm install" ("npm link" will modify files in node-modules, but "file been used" issue often occur when executed after "npm install" task).  This change still cannot solve all the issue.

src/commands/commands.ts
QuanfuXiao: Yeah, we will add commands of all hosts later.

src/taskpane/taskpane.ts
```diff
@@ -9,15 +9,78 @@ Office.onReady((info) => {
...
+  }
+  if (info.host === Office.HostType.Excel) {
+    document.getElementById("sideload-msg").style.display = "none";
+    document.getElementById("app-body").style.display = "flex";
+    document.getElementById("run").onclick = runExcel;
+  }
+  if (info.host === Office.HostType.PowerPoint) {
+    document.getElementById("sideload-msg").style.display = "none";
+    document.getElementById("app-body").style.display = "flex";
+    document.getElementById("run").onclick = runPowerPoint;
```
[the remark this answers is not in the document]
YueLi-MSFT: Yes, we hope that the template can support WXPO apps for Teams Toolkit feature.

PR #225, opened 2024-01-01
Put both types of manifest in the same branch

Rick-Kirkham:
**Change Description**:

This PR implements the user's choice of the type of manifest (xml or unified) in a single branch, so we can eliminate the "json-preview-*" branch progression and still have both types of manifests in the same repo. The idea is that when a user chooses the unified manifest option in the generator, then the generator will fetch from the yo-office branch, not the json-preview-yo-office branch, and it will pass Outlook as the first parameter to convertToSingleHost.js and also pass the string "unified" as a second parameter. The resulting project is identical to the one that is now generated if the generator uses the json-preview-yo-office branch.

Secondarily, this PR makes some of the improvements to convertToSingleHost.js that we made in the two React template repos, such as removing code that would only be needed if we were including tests in the output. 

    Yes, if the user chooses the unified manifest in Yo Office, the start, stop, and validate scripts will point to the JSON formatted manifest.

    No.

    Yes. The output that used to be only in the json-preview branch is being added to the master branch.

    No.

All tests pass locally.
Calling convertToSingleHost.js produces the same output it used to if any non-Outlook host is chosen or if Outlook is chosen, but "unified" is NOT passed as a second parameter.
It produces the desired unified manifest project if Outlook is chosen and "unified" is passed as a second parameter. The resulting project works fine.


convertToSingleHost.js
```diff
@@ -179,3 +213,15 @@ modifyProjectForSingleHost(host).catch((err) => {
   console.error(`Error: ${err instanceof Error ? err.message : err}`);
   process.exitCode = 1;
 });
+
+if ((host !== "outlook") || (manifestType !== "unified")) {
+  // Remove things that are only relevant to unified manifest
+  deleteUnifiedManifestRelatedFiles();
+  updatePackageJsonForXMLManifest();
+} else {
+    modifyProjectForUnifiedManifest().catch((err) => {
```
millerds, line 222: nit: remove extra indentation

convertToSingleHost.js
```diff
@@ -179,3 +213,15 @@ modifyProjectForSingleHost(host).catch((err) => {
   console.error(`Error: ${err instanceof Error ? err.message : err}`);
   process.exitCode = 1;
 });
+
+if ((host !== "outlook") || (manifestType !== "unified")) {
```
millerds, line 217: I would prefer using a type of "json" instead of "unified" . . . the customer facing terminology can use "unified", but the code should be clear that were looking at json.

This applies to the rest of the code as well . . . wherever we say "unified" we should say "json" instead.

millerds (changes requested):
I was thinking that the launch config, task, and webpack files are nearly identical between the xml and json versions of the template (modulo the WXP cases) and that a file modification would be better than having to maintain multiple copies of the files.  This is essentially what we do for the xml to json project conversion script.  It's a tradeoff between the simpler code in the conversion script vs. keeping the files in sync (neither of which should change all that often).  If you prefer keeping the seperate files I would put the naming difference at the end of the file name (like the xml manifest) instead of the beginning so they are together in the file tree (i.e. launch.json-manifest.json and webpack.config.json-manifest.js).

The name of the manifest file is something that generator-office needs to have in order to update it with the project name and a random guid.  First it was assumed, but then when the json manifest file was in a subfolder generator-office was modified to indicate the manifest file path as part of the config we store.  Now that we are looking at multiple manifest types in the same repo and need a parameter to be passed to the config script we need to co-ordinate what this should look like.  I'm considering: 

- go back to an assumption of the manifest file name and use the type as the extension and the argument
- passing the manifest location (which I already have) as the argument and leave the convert script to use this instead of the type (this would mean conditions are based on file path string instead of a type string).
- do some mapping somewhere (convert script or generator-office) between type and file
- use type in generator-office and pass arguments to the convert script and have it update the manifest file (it's a package command) and remove the dependency between the projects.
 
Which of these make sense to you (I'm leaning toward that last one)?

convertToSingleHost.js
Rick-Kirkham: > I would prefer using a type of "json" instead of "unified" . . . the customer facing terminology can use "unified", but the code should be clear that were looking at json.
> 
> This applies to the rest of the code as well . . . wherever we say "unified" we should say "json" instead.

OK. I'll make this change. But if we add a command line argument or option to Yo Office, it should use the term "unified" because this is the term the developers that use the tool will be familiar with.

Rick-Kirkham:
> I was thinking that the launch config, task, and webpack files are nearly identical between the xml and json versions of the template (modulo the WXP cases) and that a file modification would be better than having to maintain multiple copies of the files. This is essentially what we do for the xml to json project conversion script. It's a tradeoff between the simpler code in the conversion script vs. keeping the files in sync (neither of which should change all that often). If you prefer keeping the seperate files I would put the naming difference at the end of the file name (like the xml manifest) instead of the beginning so they are together in the file tree (i.e. launch.json-manifest.json and webpack.config.json-manifest.js).
> 

I think I can do this for the launch.json and tasks.json. Not sure if I can do it for the webpack.config.js. I've had trouble getting the string.replace() method to work when there are line breaks. 

> The name of the manifest file is something that generator-office needs to have in order to update it with the project name and a random guid. First it was assumed, but then when the json manifest file was in a subfolder generator-office was modified to indicate the manifest file path as part of the config we store. Now that we are looking at multiple manifest types in the same repo and need a parameter to be passed to the config script we need to co-ordinate what this should look like. I'm considering:
> 
> * go back to an assumption of the manifest file name and use the type as the extension and the argument
> * passing the manifest location (which I already have) as the argument and leave the convert script to use this instead of the type (this would mean conditions are based on file path string instead of a type string).
> * do some mapping somewhere (convert script or generator-office) between type and file
> * use type in generator-office and pass arguments to the convert script and have it update the manifest file (it's a package command) and remove the dependency between the projects.
> 
> Which of these make sense to you (I'm leaning toward that last one)?

I don't have a strong preference. The last one seems like it would maximize future flexibility in updating the projects.

Rick-Kirkham:
> > I was thinking that the launch config, task, and webpack files are nearly identical between the xml and json versions of the template (modulo the WXP cases) and that a file modification would be better than having to maintain multiple copies of the files. This is essentially what we do for the xml to json project conversion script. It's a tradeoff between the simpler code in the conversion script vs. keeping the files in sync (neither of which should change all that often). If you prefer keeping the seperate files I would put the naming difference at the end of the file name (like the xml manifest) instead of the beginning so they are together in the file tree (i.e. launch.json-manifest.json and webpack.config.json-manifest.js).
> 
> I think I can do this for the launch.json and tasks.json. Not sure if I can do it for the webpack.config.js. I've had trouble getting the string.replace() method to work when there are line breaks.
> 

Actually, the convertToSingleHost.js is only removing the "Debug * Tests" items from launch.json. The launch configs for all 4 of the Office hosts are still present in the generated project no matter what host was chosen. A similar point applies to the tasks.json. I noticed this before and I have a work item to fix it next spring. I'll leave it "as is" for now. I want to avoid "mission creep" in this PR.

convertToSingleHost.js
```diff
@@ -1,15 +1,14 @@
 /* global require, process, console */
 
-const convertTest = process.argv[3] === "convert-test";
 const fs = require("fs");
 const host = process.argv[2];
+const manifestType = process.argv[3];
```
millerds, line 5: Add collection of two more parameters:
- argv[4] -> projectName
- argv[5] -> appId

convertToSingleHost.js
```diff
@@ -1,15 +1,14 @@
 /* global require, process, console */
```
millerds, line 1: add "import { OfficeAddinManifest } from 'office-addin-manifest';"

convertToSingleHost.js
```diff
@@ -179,3 +241,15 @@ modifyProjectForSingleHost(host).catch((err) => {
...
+  // Remove things that are only relevant to JSON manifest
+  deleteJSONManifestRelatedFiles();
+  updatePackageJsonForXMLManifest();
+} else {
+  modifyProjectForJSONManifest().catch((err) => {
+    console.error(`Error: ${err instanceof Error ? err.message : err}`);
+    process.exitCode = 1;
+  });
+}
+
```
millerds, line 255: Add call to "await OfficeAddinManifest.modifyManifestFile(manifestPath, appId, projectName);" where the parameters are:
- manifestPath = the path to the manifest in the project (json or xml).  This is not passed into the script.
- appId = a guid for the new project.  If one is specified as an argument used that, otherwise default to the string 'random' and the method call will make one up at random
- projectName = the name to call the add-in.  Use the argument passed in if there was one.

If there was no name provided as an argument, don't call modifyManifestFile at all . . . this will make it compatible with older version of generator-office.

millerds (changes requested):
I've indicated what to add in order to handle the extra arguments for the project name and the appId.  These should be optional and the script should still work without them.  Note that in order to get the later parameters, the earlier ones in the order are required (i.e. to provide a name, the type must be indicated).

Rick-Kirkham:
> I've indicated what to add in order to handle the extra arguments for the project name and the appId. These should be optional and the script should still work without them. Note that in order to get the later parameters, the earlier ones in the order are required (i.e. to provide a name, the type must be indicated).

OK. I've made these changes. In order to use `import` statements with node.js, I had to change the file extension of convertToSingleHost to ".mjs". 

Everything seems to work. The manifest is updated as it should be regardless of whether I specify json or xml.

convertToSingleHost.mjs
```diff
@@ -0,0 +1,265 @@
+/* global require, process, console */
+
+import fs from "fs";
```
millerds, line 3: Library imports should be together and listed before the other "const" statements with a line break between.  In that case of the .js file we should have been doing that with the require statements.

convertToSingleHost.mjs
```diff
@@ -0,0 +1,265 @@
...
+const host = process.argv[2];
+const manifestType = process.argv[3];
+const projectName = process.argv[4];
+const appId = process.argv[5];
+const hosts = ["excel", "onenote", "outlook", "powerpoint", "project", "word"];
+import path from "path" ;
+import util from "util";
+import { OfficeAddinManifest } from "office-addin-manifest";
```
millerds, line 11: We could make this a require statement and keep using a '.js' extension instead of changing to '.mjs' if you prefer (my bad for giving you the import statement earlier when the file wasn't already '.mjs').  The statement would be 'const manifest = require("office-addin-manifest").'

convertToSingleHost.mjs
```diff
@@ -0,0 +1,265 @@
...
+} else {
+  manifestPath = "manifest.json";
+  modifyProjectForJSONManifest().catch((err) => {
+    console.error(`Error modifying for JSON manifest: ${err instanceof Error ? err.message : err}`);
+    process.exitCode = 1;
+  });
+};
+
+if (projectName) {
+  OfficeAddinManifest.modifyManifestFile(manifestPath, projectName, appId);
```
millerds, line 263: If you switch back to using the '.js' file extension and use the const statement, then this call would change to 'manifest.OfficeAddinManifest.modifyManifestFile(...)'.

millerds, line 263: The arguments are backwards . . . appId should be the second argument and the name should be the third.

millerds, line 263: If an appId was not passed in on the cli (i.e. the value is empty) then appId should be set to 'random' and passed in.

convertToSingleHost.mjs
```diff
@@ -0,0 +1,265 @@
+/* global require, process, console */
```
millerds, line 1: Please run prettier on this file to keep the formatting consistent.

millerds (changes requested):
It's possible to use require instead of import and keep the '.js' file extension if you prefer.  I've made comments to that effect.

If you are going to use the '.mjs' extension, then the package.json file 'convert-to-single-host' script call needs to be updated to use that extension (that is what generator-office calls).

I also noticed some problems with the call to modify the manifest . . . you might want to check that the manifest is getting modified correctly.

Rick-Kirkham:
> It's possible to use require instead of import and keep the '.js' file extension if you prefer. I've made comments to that effect.
> 
> If you are going to use the '.mjs' extension, then the package.json file 'convert-to-single-host' script call needs to be updated to use that extension (that is what generator-office calls).
> 
> I also noticed some problems with the call to modify the manifest . . . you might want to check that the manifest is getting modified correctly.

I tried `require` first, but I got an error saying that `modifyManifestFile` isn't a function. But I was using different syntax on the `require` and the actual call of the method. I'll try your suggestions. 

BTW, the manifest was being changed correctly. Although I had to put the projectName parameter before the appId parameter to get them to work correctly.

convertToSingleHost.mjs
Rick-Kirkham: > The arguments are backwards . . . appId should be the second argument and the name should be the third.

That's how I originally had it, but the manifest was updated backwards. The GUID became the name and projectName became the ID. I'll investigate.

convertToSingleHost.mjs
Rick-Kirkham: > Please run prettier on this file to keep the formatting consistent.

Running prettier on this file didn't change it. Was there something in particular you wanted changed?

millerds: When I open it in VS code there are a handful of while spaces things being flagged and a couple unnecessary parenthesis.  Not sure why we would be seeing something different.

millerds (approved):
Note that we'll need the same convert script changes in the other repos before I can update generator-office to send the extra parameters.  This is because it will stop update the manifest file when making the call.  I think it would be good for the other templates to also support json manifests . . . but that's up to you to figure out and schedule.  The convert script changes don't depend on the repos supporting json manfests, but generator-office changes depend on the convert script.

convertToSingleHost.js
```diff
@@ -1,15 +1,18 @@
 /* global require, process, console */
 
-const convertTest = process.argv[3] === "convert-test";
 const fs = require("fs");
-const host = process.argv[2];
-const hosts = ["excel", "onenote", "outlook", "powerpoint", "project", "word"];
 const path = require("path");
 const util = require("util");
+const manifest = require("office-addin-manifest");
```
YueLi-MSFT, line 6: @millerds Hi Darren, I have a question here. When will `office-addin-manifest` package be installed when node to run convertToSingleHost? Because I ran the command `npm run convert-to-single-host`, an error message `Cannot find module 'office-addin-manifest'` was reported.

millerds: Good catch.

Previously the command using this was run by the generator-office plugin which would install office-addin-manifest as part of it's setup.  However,  after this shirts the requirement shifts to the project, meaning 'npm install' needs to run first, but that is not currently how it's setup.  Currently 'yo office' will download the package and try to run the script before running 'npm install' (and the toolkit doesn't run 'npm install' as part of project generation at all).  We didn't hit this in testing because that is happening in the template repo where 'npm install' has already been run in order to do development.

@Rick-Kirkham you should be able to verify this broken behavior in using the pre-release branch if you want, but I'm pretty sure it will repro and that we should not push this to the official yo-office branch until we come up with a fix (either revert that part of the change to be in generator-office again, or do something to make sure office-addin-manifest is installed).

@YueLi-MSFT Thanks for catching this.  I just want to confirm though that you are using local changes to pull the template code from 'master' branch instead of 'yo-office' branch.  The shipped toolkit should be pulling from the 'yo-office' branch which has been vetted a little more.

Rick-Kirkham: @YueLi-MSFT Please clarify: are you running the generator (and you have found a way to point it at the master branch)? Or are you running `npm run convert-to-single-host` at the command line? You always have to run `npm install` if you are running the script at the command line.

@millerds I a little confused here. Have you actually published a new version of the generator that has the changes we talked about? At any rate I think that the generator should still run `npm install`.

millerds: I have _not_ published a new version of the generator.  My conclusions are based on my understanding of the code.  The generator will still run 'npm install' . . . the problem is the order of operations.  The convert script gets called before 'npm install' does because that's how yeoman works (finish generating a project and then call 'npm install').  

One possible fix would be force it earlier ourselves . . . but that would be taking ownership of the npm call away from yeoman and it would install things that are going to get removed by the convert script (test libraries - which would then get removed by the yeoman call to 'npm install').

YueLi is using the toolkit . . . which isn't calling yeoman, it's just doing the same type of things that generator-office does with some slight differences (like not calling 'npm install' at all).  We need to be compatible with both tools.

Rick-Kirkham: So, `npm install` would get called twice? I guess that's OK. Would it be noticeable/confusing to the end user?

millerds: If that is the fix we took then yes . . . it would be noticeable.

YueLi-MSFT: @millerds Yes, I pull the template code from master branch. I will repull code from `yo-office` branch. Thanks Darren!

YueLi-MSFT: @Rick-Kirkham Hi Rick. Yes, it is at the master branch not at the command line.

millerds: @YueLi-MSFT You are doing the right thing.  I wanted to understand the scenario you were looking at.

You are correct, this isn't in yo-office branch so it's not being used by the released 'yo office' command or teams toolkit.  To work on your WXP extensions you will need to pull from master for the time being.

millerds:
A PR has been merged that should correct this.

convertToSingleHost.js
YueLi-MSFT: @millerds Yes, I pulled code from the master branch. Currently in the toolkit code, I'm running `npm install` before `npm run convert-to-single-host`, which solves the issue. But I'm not sure the TeamsFX team(Toolkit) would approve this change, so I'd like to ask if you're going to fix it in the master branch. Thanks Darren!

millerds: https://github.com/OfficeDev/Office-Addin-TaskPane/pull/228 fixed this and should be part of master now.

PR #232, opened 2024-01-18
Fix debug tasks for hosts.

davidchesnut:

---

**Change Description**:

If you build a new project and press F5 in VS Code, you will get an error and you won't be able to debug the add-in project. This fixes the tasks so that the launch a shell with arguments correctly and debug works out of the box.

    No.

    Yes, changes are to tasks.json.

    Yes. tasks.json.

    No.

    Manual testing in progress...

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -28,8 +28,15 @@
     },
     {
       "label": "Debug: Excel Desktop",
-      "type": "npm",
-      "script": "start:desktop -- --app excel",
+      "type": "shell",
+      "command": "npm",
+      "args": [
+        "run",
+        "start:desktop",
+        "--",
+        "--app",
+        "excel",
+      ],
       "presentation": {
         "clear": true,
         "panel": "dedicated"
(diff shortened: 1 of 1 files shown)


millerds (approved):

.vscode/tasks.json
```diff
@@ -28,8 +28,15 @@
     },
     {
       "label": "Debug: Excel Desktop",
-      "type": "npm",
```
akrantz, line 31: This is really a workaround for a bug in VS Code, right? Looks fine, but really we should get VS Code to still fix the issue where arguments after `--` aren't passed correctly. Is there a VS Code bug on that?

davidchesnut: I believe it was this issue but doesn't look like this will change: https://github.com/microsoft/vscode/issues/115876

akrantz:
The response to the VS Code bug is a bit disappointing as it really is a bug they should have fixed. The workaround seems OK.

akrantz (approved):

PR #243, opened 2024-03-05
Preinstall convert requred package and keep sign in/out scripts

millerds:

---

**Change Description**:
 keeping the sign in/out script because everything has to go through the new service now.  Adding logging to the npx command to help with error that might arise for customers.

Yes.  We keep a couple in more scenarios.

No.

No.

I don't think so.

Ran tests locally and ran convert script locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "build": "webpack --mode production",
     "build:dev": "webpack --mode development",
-    "convert-to-single-host": "npm install office-addin-manifest && node convertToSingleHost.js",
+    "convert-to-single-host": "node convertToSingleHost.js",
     "dev-server": "webpack serve --mode development",
     "lint": "office-addin-lint check",
     "lint:fix": "office-addin-lint fix",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "build": "webpack --mode production",
     "build:dev": "webpack --mode development",
-    "convert-to-single-host": "node convertToSingleHost.js",
+    "convert-to-single-host": "npm install office-addin-manifest && node convertToSingleHost.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
   "scripts": {
     "build": "webpack --mode production",
     "build:dev": "webpack --mode development",
-    "convert-to-single-host": "node convertToSingleHost.js",
+    "convert-to-single-host": "npm install office-addin-manifest && node convertToSingleHost.js",
```
akrantz, line 21: This doesn't look right to me. `npm install` could update to latest and doesn't fit with package.json and package-lock.json. Isn't npm install run first?

millerds: 'npm install' is the last thing that yeoman does.  We could do some stuff to make it run in a different order by taking on the responsibility in the plugin instead of yeoman . . . but we get into situations where we are repeating stuff the convert script does (like update package.json to remove a script that updates the manifest) or yeoman does (run 'npm install' more than once in order to clean up packages that were removed from the convert script.

The other possibility is to put the install into the generator-office pipeline to fix that specific case (which is the only repro we have of the problem at the moment) . . . but that also seems odd.  I can't think of a cleaner way to doing this.

Rick-Kirkham: I don't feel qualified to review. FWIW, it is weird to call `install` in a package.json script and seems slightly less weird to move it to the pipeline. On the other hand, this line of the package.json gets deleted by the conversion script, so it won't be there to puzzle customers.

millerds: I'll put it in the CI pipeline . . . hopefully it doesn't repro in the wild.

I'd still like the logging changes though . . . that will help us if something ever does come up.

Rick-Kirkham (approved):

PR #250, opened 2024-03-15
[MetaOS] Add Addin commands files for WXPO

YueLi-MSFT:

---

**Change Description**:
This pr is to add commands files for WXPO.

   No

  No

    It will create commands related files for Word, Excel, PowerPoint and Outlook and not need to distinct the manifestType.

   No

    Describe manual testing done.


millerds (changes requested):
The basic idea is good.  However, if the actions are doing the same thing at the taskpane button we are creating duplicate code.  We should either:

- Refactor so that shared functionality is in a common file.  This would make the command and taskpane files just import the function and associate it with the button they have.
- Change the sample action to do something different (like Outlook does).  It might be more interesting for the taskpane code to do something interesting with the taskpane instead of just a button that does something.  The React template was changed to have a textbox and insert text from the box and that might be more interesting to use here and leave the command buttons with the functionality that doesn't require input.

Note that Ricky is responsible for content so his advice on what the example in the template should be is more valuable.

Rick-Kirkham:
> The basic idea is good. However, if the actions are doing the same thing at the taskpane button we are creating duplicate code. We should either:
> 
> * Refactor so that shared functionality is in a common file.  This would make the command and taskpane files just import the function and associate it with the button they have.
> * Change the sample action to do something different (like Outlook does).  It might be more interesting for the taskpane code to do something interesting with the taskpane instead of just a button that does something.  The React template was changed to have a textbox and insert text from the box and that might be more interesting to use here and leave the command buttons with the functionality that doesn't require input.
> 
> Note that Ricky is responsible for content so his advice on what the example in the template should be is more valuable.

@YueLi-MSFT Thanks for doing this. It has been on our ToDo list to create these Add-in Command functions for a long time. I like the nifty way you handled multiple hosts by moving the `associate` function into the `onReady` function. 

@millerds is right that we should have the add-in command functions do something different from what the function on the task pane does. But we're months away from public preview, so I don't think this needs to be fixed right away. I don't object to having the redundant functionality for the early private preview perio, so I'll leave it you (@YueLi-MSFT) if you want to make that change in this PR. 

Also, it is a little tricky to know what parts of the template repos are owned by my team (meaning me) and what are owned by engineering (meaning @millerds). But you can safely assume that any changes to the manifests or the files in the /src and /assets folders count as "content" and are owned by my team, so PRs that change these should include me as a reviewer.

YueLi-MSFT:
> > The basic idea is good. However, if the actions are doing the same thing at the taskpane button we are creating duplicate code. We should either:
> > 
> > * Refactor so that shared functionality is in a common file.  This would make the command and taskpane files just import the function and associate it with the button they have.
> > * Change the sample action to do something different (like Outlook does).  It might be more interesting for the taskpane code to do something interesting with the taskpane instead of just a button that does something.  The React template was changed to have a textbox and insert text from the box and that might be more interesting to use here and leave the command buttons with the functionality that doesn't require input.
> > 
> > Note that Ricky is responsible for content so his advice on what the example in the template should be is more valuable.
> 
> @YueLi-MSFT Thanks for doing this. It has been on our ToDo list to create these Add-in Command functions for a long time. I like the nifty way you handled multiple hosts by moving the `associate` function into the `onReady` function.
> 
> @millerds is right that we should have the add-in command functions do something different from what the function on the task pane does. But we're months away from public preview, so I don't think this needs to be fixed right away. I don't object to having the redundant functionality for the early private preview perio, so I'll leave it you (@YueLi-MSFT) if you want to make that change in this PR.
> 
> Also, it is a little tricky to know what parts of the template repos are owned by my team (meaning me) and what are owned by engineering (meaning @millerds). But you can safely assume that any changes to the manifests or the files in the /src and /assets folders count as "content" and are owned by my team, so PRs that change these should include me as a reviewer.

I am writing to thank you for your valuable insights about taskpane and command. I agree with @millerds's suggestion to refactor the code before public review. I will work on this as soon as possible and create a new PR accordingly.
@Rick-Kirkham Wow, that's great! I will do it for private review in the json-wxpo-preview branch.  I hope you can approve this PR after reviewing the changes. Thank you for your time and collaboration!

Rick-Kirkham:
> > > The basic idea is good. However, if the actions are doing the same thing at the taskpane button we are creating duplicate code. We should either:
> > > 
> > > * Refactor so that shared functionality is in a common file.  This would make the command and taskpane files just import the function and associate it with the button they have.
> > > * Change the sample action to do something different (like Outlook does).  It might be more interesting for the taskpane code to do something interesting with the taskpane instead of just a button that does something.  The React template was changed to have a textbox and insert text from the box and that might be more interesting to use here and leave the command buttons with the functionality that doesn't require input.
> > > 
> > > Note that Ricky is responsible for content so his advice on what the example in the template should be is more valuable.
> > 
> > 
> > @YueLi-MSFT Thanks for doing this. It has been on our ToDo list to create these Add-in Command functions for a long time. I like the nifty way you handled multiple hosts by moving the `associate` function into the `onReady` function.
> > @millerds is right that we should have the add-in command functions do something different from what the function on the task pane does. But we're months away from public preview, so I don't think this needs to be fixed right away. I don't object to having the redundant functionality for the early private preview perio, so I'll leave it you (@YueLi-MSFT) if you want to make that change in this PR.
> > Also, it is a little tricky to know what parts of the template repos are owned by my team (meaning me) and what are owned by engineering (meaning @millerds). But you can safely assume that any changes to the manifests or the files in the /src and /assets folders count as "content" and are owned by my team, so PRs that change these should include me as a reviewer.
> 
> I am writing to thank you for your valuable insights about taskpane and command. I agree with @millerds's suggestion to refactor the code before public review. I will work on this as soon as possible and create a new PR accordingly.
> 
> @Rick-Kirkham Wow, that's great! I will do it for private review in the json-wxpo-preview branch. I hope you can approve this PR after reviewing the changes. Thank you for your time and collaboration!

@YueLi-MSFT , I want to be sure we are on the same page. Refactoring as @millerds suggested in his first bullet is something you can do for the PRIVATE preview if you want to remove redundant code. But for PUBLIC preview we must have function commands that do something different from what is done by the button or link on the task pane. Do we have your agreement about this?

YueLi-MSFT:
> > > > The basic idea is good. However, if the actions are doing the same thing at the taskpane button we are creating duplicate code. We should either:
> > > > 
> > > > * Refactor so that shared functionality is in a common file.  This would make the command and taskpane files just import the function and associate it with the button they have.
> > > > * Change the sample action to do something different (like Outlook does).  It might be more interesting for the taskpane code to do something interesting with the taskpane instead of just a button that does something.  The React template was changed to have a textbox and insert text from the box and that might be more interesting to use here and leave the command buttons with the functionality that doesn't require input.
> > > > 
> > > > Note that Ricky is responsible for content so his advice on what the example in the template should be is more valuable.
> > > 
> > > 
> > > @YueLi-MSFT Thanks for doing this. It has been on our ToDo list to create these Add-in Command functions for a long time. I like the nifty way you handled multiple hosts by moving the `associate` function into the `onReady` function.
> > > @millerds is right that we should have the add-in command functions do something different from what the function on the task pane does. But we're months away from public preview, so I don't think this needs to be fixed right away. I don't object to having the redundant functionality for the early private preview perio, so I'll leave it you (@YueLi-MSFT) if you want to make that change in this PR.
> > > Also, it is a little tricky to know what parts of the template repos are owned by my team (meaning me) and what are owned by engineering (meaning @millerds). But you can safely assume that any changes to the manifests or the files in the /src and /assets folders count as "content" and are owned by my team, so PRs that change these should include me as a reviewer.
> > 
> > 
> > I am writing to thank you for your valuable insights about taskpane and command. I agree with @millerds's suggestion to refactor the code before public review. I will work on this as soon as possible and create a new PR accordingly.
> > @Rick-Kirkham Wow, that's great! I will do it for private review in the json-wxpo-preview branch. I hope you can approve this PR after reviewing the changes. Thank you for your time and collaboration!
> 
> @YueLi-MSFT , I want to be sure we are on the same page. Refactoring as @millerds suggested in his first bullet is something you can do for the PRIVATE preview if you want to remove redundant code. But for PUBLIC preview we must have function commands that do something different from what is done by the button or link on the task pane. Do we have your agreement about this?

Of course. That make sense.

Rick-Kirkham (approved):

millerds:
Aside from making sure the tests pass . . . what the plan for the duplicate code (between actions and taskpane buttons)?

package.json
```diff
@@ -35,6 +35,7 @@
     "watch": "webpack --mode development --watch"
   },
   "dependencies": {
+    "@microsoft/teamsapp-cli": "^3.0.0",
```
millerds, line 38: Why is this explicitly needed?  There should not be anything in the template that has a direct dependency on this (only indirectly through other packages that use it).

millerds (changes requested):

.vscode/launch.json
```diff
@@ -54,7 +54,7 @@
     },
     {
       "name": "Excel Desktop (Edge Legacy)",
-      "type": "office-addin",
+      "type": "msedge",
```
millerds, line 57: This does not work.  The built in 'msedge' debugger in VSC is not able to attach to host running the legacy edge webview . . . that is why the custom debugger was created and referenced here.

millerds (changes requested):

Rick-Kirkham (changes requested):
I have some concerns: 

1. Remove the "displayName" property from the manifest. It is only used for keyboard shortcuts which are not part of this template.
2. The functions "actionExcel" , "actionWord", etc. should be given ordinary JavaScript function names based on what they do. For example, "highlightRange"or "insertText". 
3. Having multiple values in the "extensions.requirements.scopes" array only makes sense if the template is for developers who choose to create an add-in for multiple Office applications. But they will be choosing a particular Office application. Having a manifest designed for multiple Office applications will mislead and confuse. I think we should treat the JSON manifest the same way that we treat the XML manifest: have a different manifest for each application and then the convertToSingleHost.js can remove the manifests for the applications that the user did not choose. 
4. When we've done #3, then we don't have to have the vague term "action" as the value of the "actions.id" in the CommandsRuntime. We can give each action a more descriptive name. 
5. The "requirements" property in the second TaskPaneRuntime object is in the wrong place. It should be in the CommandsRuntime object.

manifest.json
```diff
@@ -44,11 +44,7 @@
   "extensions": [
     {
       "requirements": {
-        "scopes": ["mail", "workbook", "document", "presentation"],
-        "capabilities": [
-          { "name": "Mailbox", "minVersion": "1.3" }, 
-          { "name": "AddinCommands", "minVersion": "1.1" }
-        ]
+        "scopes": ["mail", "workbook", "document", "presentation"]
```
Rick-Kirkham, line 47: Having more than one is not a common scenario and is not consistent with the UI in either Yo Office or Teams Toolkit in which the user will select one Office application.

src/commands/excel.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (info.host === Office.HostType.Excel) {
+    // Register the function with Office.
+    Office.actions.associate("action", actionExcel);
+  }
+});
+/**
+ * Shows a notification when the add-in command is executed.
+ * @param event
+ */
+async function actionExcel(event: Office.AddinCommands.Event) {
```
Rick-Kirkham, line 18:
```suggestion
async function setRangeColor(event: Office.AddinCommands.Event) {
```

src/commands/outlook.ts
```diff
@@ -0,0 +1,32 @@
...
+    // Register the function with Office.
+    Office.actions.associate("action", actionOutlook);
+  }
+});
+
+/**
+ * Shows a notification when the add-in command is executed.
+ * @param event
+ */
+function actionOutlook(event: Office.AddinCommands.Event) {
```
Rick-Kirkham, line 19:
```suggestion
function setNotification(event: Office.AddinCommands.Event) {
```

src/commands/excel.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+
+Office.onReady((info) => {
+  if (info.host === Office.HostType.Excel) {
+    // Register the function with Office.
+    Office.actions.associate("action", actionExcel);
```
Rick-Kirkham, line 11:
```suggestion
    Office.actions.associate("action", setRangeColor);
```

src/commands/outlook.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+
+Office.onReady((info) => {
+  if (info.host === Office.HostType.Outlook) {
+    // Register the function with Office.
+    Office.actions.associate("action", actionOutlook);
```
Rick-Kirkham, line 11:
```suggestion
    Office.actions.associate("action", setNotification);
```

src/commands/powerpoint.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+
+Office.onReady((info) => {
+  if (info.host === Office.HostType.PowerPoint) {
+    // Register the function with Office.
+    Office.actions.associate("action", actionPowerPoint);
```
Rick-Kirkham, line 11:
```suggestion
    Office.actions.associate("action", insertText);
```

src/commands/powerpoint.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+});
+
+/**
+ * Shows a notification when the add-in command is executed.
+ * @param event
+ */
+async function actionPowerPoint(event: Office.AddinCommands.Event) {
```
Rick-Kirkham, line 19:
```suggestion
async function insertText(event: Office.AddinCommands.Event) {
```

src/commands/word.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+
+Office.onReady((info) => {
+  if (info.host === Office.HostType.Word) {
+    // Register the function with Office.
+    Office.actions.associate("action", actionWord);
```
Rick-Kirkham, line 11:
```suggestion
    Office.actions.associate("action", insertBlueParagraph);
```

src/commands/word.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+});
+
+/**
+ * Shows a notification when the add-in command is executed.
+ * @param event
+ */
+async function actionWord(event: Office.AddinCommands.Event) {
```
Rick-Kirkham, line 19:
```suggestion
async function insertBlueParagraph(event: Office.AddinCommands.Event) {
```

Rick-Kirkham (changes requested):

akrantz:
I just wrote a sample using contextual tabs where you need a function to execute for the ribbon button clicked. I put this in actions.ts and used `action` as a prefix, e.g. actionSignIn. I think that's better than just generic function names because it reinforces that these are specifically for executing some actions.  @Rick-Kirkham

manifest.json
akrantz: We need to consider scenarios where add-ins target more than one host app. I am working with a customer now that has a scenario where the add-in runs in two different Office hosts. While the common case is only one, we need to allow for more flexibility to support more.

src/commands/outlook.ts
akrantz: Seems like the name here should not be the generic "action" but more specific. I personally think it should be the same for the string as the function name, e.g. Office.actions.associate("actionSignIn", actionSignIn);

Rick-Kirkham: > Seems like the name here should not be the generic "action" but more specific. I personally think it should be the same for the string as the function name, e.g. `Office.actions.associate("actionSignIn", actionSignIn);`
> 
> For this case, why not use `actionSetNotification` for both?

That gets to an issue I raised in my other comment. As currently structured in the PR, the template is for multiple Office applications. But this means there is a single CommandsRuntime runtime object that must apply to all the possible hosts. So, the "actions.id" in that object has to be a generic name that can be used in any of the Office applications. And the first parameter of the `associate` method has to match that name. We could do as you suggest, if we have separate manifests for each host.

manifest.json
Rick-Kirkham: > We need to consider scenarios where add-ins target more than one host app. I am working with a customer now that has a scenario where the add-in runs in two different Office hosts. While the common case is only one, we need to allow for more flexibility to support more.

I don't agree that we need to support multiple hosts _in the template_. The templates are for starter projects. The end user will have to make many changes as they develop their add-in. Adding additional Office hosts is just one more such change.

Rick-Kirkham:
> I just wrote a sample using contextual tabs where you need a function to execute for the ribbon button clicked. I put this in actions.ts and used `action` as a prefix, e.g. actionSignIn. I think that's better than just generic function names because it reinforces that these are specifically for executing some actions. @Rick-Kirkham
> 
> image

But will it always be used for executing an action? Couldn't the same file and function be included in the bundle for a task pane and invoked in the UI of the task pane?

millerds (changes requested):
Note that the PR and branch are old enough that they out to incorporate the changes made in the mater branch in order to keep in sync.

manifest.json
Rick-Kirkham: > One of the goals of the WXP work with the new manifest is to show case the improved support for multiple hosts in one project and you see that evident in other parts of this branch. There is a "manifest.json" that is for multiple hosts and a manifest..json for the individual case.

@millerds So are you envisioning that Yo Office will give users the choice of hosts, but also the option of choosing a combined WXPO project? I can live with that, but: 

1. I wonder why we're expending labor on an option that historically has been unpopular.
2. If the user takes the WXPO option, but they want to remove one or more of those hosts, they'll have to edit the manifest anyway, just as they do if they take a one-host option and then edit the manifest to add another host. The multiple host option is labor-saving only if WXPO is exactly the combination that they want.

Also, with respect to the "improved support for multiple hosts in one project," it's always been the case that devs can include WXP (and, for that matter, OneNote and Project too) into a single add-in. (Very few devs wanted to do this.) So, the improvement is that you can now add Outlook into the mix, which doesn't seem like all that much of an improvement. In fact, since OneNote and Project can _not_ be included in the add-in with the new manifest, maybe it's not a net improvement at all.

millerds: Right now I believe the toolkit (which also pulls from this repo for its template) changes they are making currently the only case that does the multi-host project generation.  This should be easily added to 'yo office' as well.

I'm not aware of any data that says one way or another that a mutli-host project generation is desirable or not (it would be interesting to see).  I have heard anecdotally of existing add-ins that have multiple apps in the app store because they can't combine Outlook with the others and that is one problem the new manifest aims to solve so why not have the tooling support that.  True that if they don't want everything but want more than one host they have to remove stuff.  They could also start from a single case and add it (which is still an option).  Whether manually adding or removing pieces is easier seems like a matter of opinion.

I personally envision an update to generator-office and the template repo that allows for a choice of multiple hosts and multiple features in a generated project so that we can create a greater variety of starting points (we periodically get github issues along these lines) with less repos to maintain.  Having a starting point with everything and then surgically removing things in the convert script is easier than trying to programmatically add stuff.

Rick-Kirkham: > Right now I believe the toolkit (which also pulls from this repo for its template) changes they are making currently the only case that does the multi-host project generation. This should be easily added to 'yo office' as well.
> 

I agree with the last paragraph. Way last fall when we were talking about whether to include add-in commands in the task pane templates, it was unanimously agreed that someday we want to have a "menu" system in which the developer selects the features and hosts they want to include in their add-in. 

I can't quite follow your first paragraph about the toolkit. Could you clarify?

millerds: > I can't quite follow your first paragraph about the toolkit. Could you clarify?

When the tool kit generates a project for office add-in they are using this repo for the project code in the same way that 'yo office' does, but they pull from different branches.  They have changes that based on updated UI options for project generation provide the "wxpo" argument to the convert script in this branch that keeps the code for all hosts.  So far that has not been added to the 'yo office' options in the generator-office repo, but potentially could be.

.vscode/launch.json
hermanwenhe: Since the debugger type "office-addin" is deprecated along with the add-in debugger (will be replaced by Dev Kit but not with this debugger type) and documentation, do you suggest that we remove all the Edge Legacy profiles?

manifest.json
hermanwenhe: Looks like we all agree to keep this line with multiple hosts and Darren will later figure out a way to remove other hosts in the conversion script in future?

hermanwenhe:
> Note that the PR and branch are old enough that they out to incorporate the changes made in the mater branch in order to keep in sync.

Agree that this PR and branch is too old. Can you sync the target branch [OfficeDev:json-wxpo-preview](https://github.com/OfficeDev/Office-Addin-TaskPane/tree/json-wxpo-preview) with master branch first?

millerds:
> > Note that the PR and branch are old enough that they out to incorporate the changes made in the mater branch in order to keep in sync.
> 
> Agree that this PR and branch is too old. Can you sync the target branch [OfficeDev:json-wxpo-preview](https://github.com/OfficeDev/Office-Addin-TaskPane/tree/json-wxpo-preview) with master branch first? Or do you suggest to change PR target to master branch?

We do not want to target master until we're ready to GA.  json-wxpo-preview is your teams' responsibility and so I wasn't planning in updating it myself (just reviewing the changes).

.vscode/launch.json
millerds: The tricky part is that we still have version of Office that we support that use the legacy web view and even the IE web view.  Developers making add-ins that need to be supported in those versions will want to be able to figure out problems with those add-ins.

The debugger type was deprecated only because we couldn't make security updates for it and necessary libraries were also no longer getting updated.  The debugger was still available in plugin store to use so it something was still available.  I think when we made the decision to repurpose that plugin entry with the DevKit we essentially made the decision to take away the only debug experience we have for that type of view.  

If the debugger plugin is pulled from the VS Code store, then it doesn't make sense to keep these debug configurations.

manifest.json
millerds: That is my understanding.

src/commands/excel.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+
+Office.onReady((info) => {
+  if (info.host === Office.HostType.Excel) {
+    // Register the function with Office.
+    Office.actions.associate("action", setRangeColorInExcel);
```
millerds, line 11: Even though the host specific method names changed they are all still getting associated with the same "action" command being referenced in the manifest.  You need to have an intermediate method that determines which host specific method to call based on which host is running.  (see the react template for such an example).

millerds (changes requested):

millerds:
Also note there are linter errors being flagged in the pipeline.

.vscode/launch.json
hermanwenhe: I think there is an alternative way of debugging legacy webviews. https://learn.microsoft.com/en-us/office/dev/add-ins/testing/debug-add-ins-using-devtools-edge-legacy  Content writer's have updated this document to remove the debugger section(because it's deprecated). So I think technical there is not and should not be any traffic to install the deprecated debugger.

millerds: That makes sense.

src/commands/excel.ts
hermanwenhe: @millerds  Resolved, pls check.

.vscode/launch.json
hermanwenhe: Remove legacy profiles in latest commit.

hermanwenhe:
I've updated a new version of methods naming and dynamic associating the action method with reference to React template. Can you share thoughts on this? @Rick-Kirkham @millerds @akrantz

hermanwenhe:
@millerds Looks like there is a pipeline permission issue.

src/commands/excel.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+/* global Excel console */
```
millerds, line 7: These should be combined to one comment.

src/commands/powerpoint.ts
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+/* global PowerPoint console */
```
millerds, line 7: These should be combined to one comment.

src/commands/word.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT license.
+ * See LICENSE in the project root for license information.
+ */
+
+/* global Office */
+/* global Word console */
```
millerds, line 7: These should be combined to one comment.

src/commands/commands.ts
```diff
@@ -1,32 +1,27 @@
...
+      break;
+    case Office.HostType.PowerPoint:
+      Office.actions.associate("action", insertTextInPowerPoint);
+      break;
+    case Office.HostType.Outlook:
+      Office.actions.associate("action", setNotificationInOutlook);
+      break;
+    default: {
+      throw new Error(`${info.host} not supported.`);
+    }
```
millerds, line 25: This is the type of switch statement we need for the multi host scenario, but in the case of a single host we don't seem to be cleaning up cases.  Either we need to have convertToSingle host modify this file to remove host case statements and import statements that are not supported in the generated project, or move them to the individual host files (each having their own onReady call with an "if host equals" statement) so it gets cleaned up with those files (this file would just be import statements of the files, which also need to get cleaned up).

src/commands/commands.ts
```diff
@@ -1,32 +1,27 @@
...
 
 /* global Office */
 
-Office.onReady(() => {
-  // If needed, Office.js is ready to be called.
+// Register the add-in commands with the Office host application.
+Office.onReady(async (info) => {
+  switch (info.host) {
+    case Office.HostType.Word:
+      Office.actions.associate("action", insertBlueParagraphInWord);
```
millerds, line 12: I just want to make sure that these "Office.actions.associate" methods work well in the "onReady" call. I've Only ever seen them used in the root of a file rather with in another office.js method.  I don't see why it wouldn't work; I just want to make sure it was tested.

millerds (changes requested):
We need to make sure the changes work in the multi host scenario and the single host scenario . . . including cleaning up stuff belonging to hosts that won't be supported in the generated project.

millerds:
> @millerds Looks like there is a pipeline permission issue.

The master branch seems to pass so I'm not sure what is up.  The part that is failing it from the pipeline template we are being told to use as part of security measures.  When I merged from master these pipeline changes were added to the branch.

src/commands/excel.ts
hermanwenhe: Fixed

src/commands/commands.ts
hermanwenhe: Just pushed a new commit. Pls check.

Basically it will replace the whole file content with the right onReady statement for single host and delete other useless host commands file.

src/commands/commands.ts
hermanwenhe: If the pipeline works well can it cover the testing work? I just want to make sure that if testing by hand is the right way to do validation.

millerds: Any change requires some validation of the change outside depending on the pipeline (which is not comprehensive).  In this case you are add-in command functionality to the template and that is not covered by tests (even if the pipeline was working for forked builds).  A simple check that you can run the add-in in each host and click the command button at it works is all that is needed.  Adding tests to cover this would be even better, but since we haven't been great at test coverage before I wouldn't all of the sudden force it now.

src/commands/commands.ts
millerds: This is not the methodology we've adopted for these repos.  Doing it this way means is something changes there is a file that gets updated and code in the convert script that also needs to get updated and the two need to stay in sync.  We will always have some of this, but we try to keep this to a minimum by using files that get overwritten and/or deleted.  This also makes it easier for minor updates to be done by less technical people who can just look at it as content most of the time.

For examples of this see the earlier parts of the convertProjectToSingleHost method where we are the taskpane code files to keep and just updating the import statements of the main file.

millerds (changes requested):
We should continue the convertToSingleHost pattern of managing files instead of code.

src/commands/commands.ts
hermanwenhe: Understood. I'll do the validation of this part.

hermanwenhe:
> We should continue the convertToSingleHost pattern of managing files instead of code.

Done. Pls check.

src/commands/commands.ts
hermanwenhe: Done.

Rick-Kirkham (approved):

hermanwenhe:
Manually tested on all hosts.

millerds:
@hermanwenhe I accidentally merged these changes when I still has some comments I'd like addressed.  I will note them here and suggest they be considered in a future PR and before we try and merge with master when GA comes along.

millerds (commented):
Please consider these in a future update (changes were already merged).

PR #272, opened 2024-09-06
Fix: delete duplicate hosts declaration

hermanwenhe:

---

**Change Description**:

    Describe what the PR is about. 

    No

    No

    No

    No

    Describe manual testing done.

### DIFF
--- a/convertToSingleHost.js
+++ b/convertToSingleHost.js
@@ -22,7 +22,6 @@ const host = process.argv[2];
 const manifestType = process.argv[3] || "xml";
 const projectName = process.argv[4];
 let appId = process.argv[5];
-const hosts = ["excel", "onenote", "outlook", "powerpoint", "project", "word"];
 const testPackages = [
   "@types/mocha",
   "@types/node",


convertToSingleHost.js
```diff
@@ -165,7 +164,7 @@ async function updateLaunchJsonFile(host) {
   const launchJsonContent = await readFileAsync(launchJson, "utf8");
   let content = JSON.parse(launchJsonContent);
   content.configurations = content.configurations.filter(function (config) {
-    return config.name.startsWith(getHostName(host));
+    return host == "wxpo" ? true : config.name.startsWith(getHostName(host));
```
millerds, line 167: I believe this would leave all the launch configs in there in the case of "wxpo", including the test debugging cases which should get removed.

This should be change use an array like is done in convertProjectToSingleHost.  Eventually I think we need to convert to use an array for the target host all over this file instead of assuming just one.  This would help avoid having to special case 'wxpo' separately everywhere.

millerds (changes requested):
I don't think that has quit the behavior we want.

convertToSingleHost.js
hermanwenhe: Done

millerds (approved):

hermanwenhe:
@millerds Please help merge

PR #362, opened 2026-07-09
Fix unit tests on Node 24 and prevent E2E runs from hanging in CI

AlexJerabek:
## Problem

Recent PRs fail on the **`Office-Addin-TaskPane (stage Windows_Latest)`** check. Two distinct issues:

1. **Unit tests break on Node 24.** Node 24 enables native TypeScript type-stripping by default. It intercepts the `.ts` test files before `ts-node/register` and loads them as ES modules, so `require(...)` in the tests throws `ReferenceError: require is not defined` (Excel/PowerPoint/Word "Run").
2. **E2E stage hangs until the job timeout.** `test/end-to-end/ui-test.ts` sideloads desktop Excel/PowerPoint/Word and every hook/test used `this.timeout(0)`. When a sideload stalls, the run blocks until the CI job's ~60-min cap and is **cancelled** — which matches the identical ~1h cancellations seen across unrelated dependency-bump PRs.

## Changes

- **`.mocharc.json`** — adds `--no-experimental-strip-types` so ts-node handles `.ts` as CommonJS on Node 24. Applies to both unit and E2E mocha runs.
- **`test/end-to-end/ui-test.ts`** — replaces `this.timeout(0)` with a finite `testTimeout` (5 min) on the sideload/result hooks so a hang fails fast instead of burning the whole job.
- **`.azure-devops/full-pipeline.yml`** — pins the pipeline to Node 22.x (`NodeTool@0`) to stop pool-image drift and guarantee a Node that supports the mocha flag.
- **`package.json`** — bumps `engines.node` floor to `>=22` (Node 20 is EOL; the mocha flag needs ≥22.6).
- **`convertToSingleHost.js`** — removes the now-orphaned `.mocharc.json` during single-host conversion.

## Validation

Ran locally on **Node v24.12.0**:
- `npm run lint` ✅
- `npm run build` ✅
- `npm run test:unit` ✅ (3 passing — previously 3 failing with the require error)

⚠️ **`npm run test:e2e` was not run locally** — it sideloads desktop Office and requires a machine with Office installed. The `testTimeout` value (5 min) may need tuning on a real CI agent; reviewers with an Office-equipped environment should validate the E2E behavior.

Opened as a **draft** pending E2E validation.

### DIFF
--- a/.azure-devops/full-pipeline.yml
+++ b/.azure-devops/full-pipeline.yml
@@ -19,6 +19,10 @@ extends:
       jobs:
       - job: Windows_Latest
         steps:
+        - task: NodeTool@0
+          displayName: 'Use Node 22.x'
+          inputs:
+            versionSpec: '22.x'
         - template: /.azure-devops/install.yml@self
         - template: /.azure-devops/lint.yml@self
         - template: /.azure-devops/build.yml@self
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "dev_server_port": 3000
   },
   "engines": {
-    "node": ">=20 <25",
+    "node": ">=22 <25",
     "npm": ">=9 <12"
   },
   "scripts": {
--- a/test/end-to-end/ui-test.ts
+++ b/test/end-to-end/ui-test.ts
@@ -12,6 +12,9 @@ import * as testHelpers from "./src/test-helpers";
 const hosts = ["Excel", "PowerPoint", "Word"];
 const manifestPath = path.resolve(`${process.cwd()}/test/end-to-end/test-manifest.xml`);
 const testServerPort: number = 4201;
+// Finite timeout so a stuck sideload fails within minutes instead of hanging
+// until the CI job's timeout cancels the whole run.
+const testTimeout: number = 5 * 60 * 1000;
 
 hosts.forEach(function (host) {
   const testServer = new officeAddinTestServer.TestServer(testServerPort);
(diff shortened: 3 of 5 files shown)


.azure-devops/full-pipeline.yml
```diff
@@ -19,6 +19,10 @@ extends:
       jobs:
       - job: Windows_Latest
         steps:
+        - task: NodeTool@0
+          displayName: 'Use Node 22.x'
+          inputs:
+            versionSpec: '22.x'
```
millerds, line 25: Note that by pegging the node version in the pipeline yaml you take on the responsibility of keeping it up to date instead of letting the image do that. You do gain more control of the pipeline details if that is more important (it's a trade off.

package.json
```diff
@@ -12,7 +12,7 @@
     "dev_server_port": 3000
   },
   "engines": {
-    "node": ">=20 <25",
+    "node": ">=22 <25",
```
millerds, line 15: Note that v26 is being produced and will become the default LTS version in the fall . . . if you wanted to do a quick check that the template works with that version you could make that change here now as well.  (or it would be done later).

test/end-to-end/ui-test.ts
```diff
@@ -39,7 +42,7 @@ hosts.forEach(function (host) {
     }),
```
millerds, line 42: I think the ',' should be a ';' . . . probably an overlooked mistake that has been there for quite some time (as it the extra indentation of the "describe" statement that comes next).

millerds:
I'm not sure why the type stripping would cause a timeout issues.  It seems like an OK thing to change, but I'm not sure how it helps.

AlexJerabek:
> I'm not sure why the type stripping would cause a timeout issues. It seems like an OK thing to change, but I'm not sure how it helps.

@millerds, this was just investigation. Seems like the root cause of the test failures has something to do with an environmental change in the test pool. How would we go about investigating that?

